PR #74, opened 2020-03-06
Update links to oneCCL repo

outoftardis:
oneCCL moved to another repository. The redirects from the old repo are enabled, but the links should be updated at some point.


rscohn2:
Merged. Thanks for the update.

PR #81, opened 2020-03-16
[WIP]: Skeleton for oneDAL algorithm

rlnx:
**Changes:**
- Updated structure of oneDAL spec
- Started working on spec skeleton for K-Means
  - Added high-level algorithm description
  - Added K-Means API definitions

screencapture-file-C-locrepo-oneapi-spec-build-html-elements-oneDAL-source-algorithms-clustering-kmeans-html-2020-03-16-17_57_37


source/elements/oneDAL/source/algorithms/index.rst
```diff
@@ -1,14 +1,8 @@
-============
- Algorithms
-============
+==========
+Algorithms
+==========
 
 .. toctree::
```
outoftardis: I'd suggest you use here a maxdepth limit for toc, otherwise you get a table of content that is too messy

```
.. toctree::
   :maxdepth: 2
```
toc-2max
toc-nolimit

outoftardis:
Just a suggestion:
if you write `.. highlight:: cpp` at the top of the current rst doc, you can then just use `::` instead of `.. code-block:: c++` in that document

andreynv:
Quick comments from my side:
- we need to stabilize terminology - feature vectors/observations/points. The def should be introduced in the common part and used consistently across the specs 

- the level of algorithm description - how many details we want/can give, reference to bibliography, which precisely method we specify, so spec implementer would easily do it based on the info in the specs/additional info in the literature

- related to the comment to maximum toc: we need to estimate how it fits to the skeleton of the specs created by  Nikolay and minimize nested levels as much as possible (it is similar to namespace topic) and stabilize classification of the algorithms/level in toc. I assume that for each algorithm we would have API/input/...some kind of duplication. One option is to have a section describing this specifics for all algorithms, then specific algorithm would give the info in the flat order.

- duplication of namespace in input/refer etc. Should we mention the namespace one time only (and the common section in the previous item would describe this aspect as well)

- we need paragraphs on pre-/post-conditions

- description of the classes/their methods/specific parameters/default values/templates - how do we want to systematically cover/describe them?

- how do we want to highlight specifics of batch/online/distributed in the skeleton?

- crossreferences (e.g., to table description or so)-to blank sections now

- usage example - would it be duplication of the lib examples? Probably, yes as those are specs. Then the question is when we re-use the specs in regular dal product doc-how it would fit/be embedded into c++/java/etc specifics with examples as the files?

- minor: we have both named/unnamed method parameters - how do we want to converge?

outoftardis:
@andreynv 
> how do we want to highlight specifics of batch/online/distributed in the skeleton?

what about tabbed view? the same way we did it with lasso/elastic or linear/ridge regressions in Developer Guide.

outoftardis:
> crossreferences (e.g., to table description or so)-to blank sections now

let me know if you need help with this in rst syntax

andreynv:
@outoftardis, looks interesting, make sense to analyze using 1-2 complex examples

rlnx:
@andreynv 
> related to the comment to maximum toc: we need to estimate how it fits to the skeleton of the specs created by Nikolay and minimize nested levels as much as possible (it is similar to namespace topic) and stabilize classification of the algorithms/level in toc. I assume that for each algorithm we would have API/input/...some kind of duplication. One option is to have a section describing this specifics for all algorithms, then specific algorithm would give the info in the flat order.

I've updated skeleton/toc in this PR (see attached image). There is info about descriptors concept and operations (train/infer) in "Algorithm anatomy" section. Specific algorithms give description of their input and results:
- Descriptor
- Model
- Training
   - Input
   - Result
- Inference
  - Input
  - Result

This structure will not be part of toc, so for the reader it looks like "flat" and shown in one page.

source/elements/oneDAL/source/algorithms/index.rst
rlnx: Fixed by c7866e3

andreynv:
> I've updated skeleton/toc in this PR (see attached image). There is info about descriptors concept and operations (train/infer) in "Algorithm anatomy" section. Specific algorithms give description of their input and results:

it looks more solid. The observation is about mixing algorithm anatomy, data management and algorithms again-probably a different order that highlights logically connected parts together would be better

PR #83, opened 2020-03-20
Update oneVPL elements.

pengtu:
Please pull the updated oneVPL components.  Note that with this change, graphviz is required to generate the state transition diagrams in oneVPL.  I only updated the 'extensions' in the common conf.py.  I don't have experience updating the 'requirements.txt' and the build environment setting instructions.  Please help updating them after pulling into the master.


rscohn2:
I will work in your branch to get the build working and then we can merge into master.

pengtu:
I fixed the requirement.txt.  We still need to install 'graphviz' to the build environment to fix the missing 'dot' problem.

PR #84, opened 2020-03-20
Add detailed K-Means API description

rlnx:
Review the changes here: https://rlnx.github.io/oneapi-spec/elements/oneDAL/source/algorithms/clustering/kmeans.html


source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
+   Invariants
+      | ``cluster_count > 0``
+
+max_iterations ``int64_t`` ``default = 100``
+   The maximum number of iterations.
+
+   Invariants
+      | ``max_iterations > 0``
+
+accuracy_threshold ``double`` ``default = 0.0``
```
andreynv: it would be easier for implementer consume those parameters (I also keep in mind other, more complex algorithms), if they are introduced/described somewhere in the beginning
also, we need to specify, what this threshold means, that is what stop criteria, what would happen, if threshold is set to zero

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
+   Invariants
+      | ``centroids.is_empty == false``
+
+cluster_count ``int64_t`` ``[read-only]``
+   Number of clusters in the trained model.
 
-Training
---------
+   Invariants
+      | ``cluster_count == centroids.row_count``
```
andreynv: I assume we use 'centroids' as the name for the table with centroids. If yes, we need to define this name earlier

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
+::
 
    class model {
    public:
       const table& get_centroids() const;
+      int64_t get_cluster_count() const;
    };
 
+centroids ``table`` ``[read-only]``
+   :math:`[\mathrm{cluster\_count} \times \mathrm{feature\_count}]` table with the
```
andreynv: table with k centroids of dimension p each?
Also, we specify the layout of the table, each row contains one centroid. I assume that generally we can store the centroids differently, for perf reasons

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
+::
 
    namespace method {
-   struct lloyd {};
-   using by_default = lloyd;
+      struct lloyd {};
+      using by_default = lloyd;
    } // namespace method
 
    template <typename Float = float,
```
andreynv: we have to describe those parameters, float/method as well, probably one time in the single place and refer from the algorithm to that section

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
 ~~~~~
-.. code-block:: c++
-
-   namespace onedal::kmeans {
+::
 
    class train_input {
    public:
       train_input();
       train_input(const table& data);
```
andreynv: here we use parameter names, to make clear which table means what? If yes, we will need a simple rule

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
+      const table& get_initial_centroids() const;
+
       train_input& set_data(const table&);
       train_input& set_initial_centroids(const table&);
    };
 
-   } // namespace onedal::kmeans
+data ``table`` ``default = table()``
+   :math:`[\mathrm{observation\_count} \times \mathrm{feature\_count}]` table with the data
+   to be clustered, where each row stores one observation.
```
andreynv: feature vector?
table of n feature vectors of dimension p ot correlate with the problem statement in the beginning?

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
 
+initial_centroids ``table`` ``default = table()``
+   :math:`[\mathrm{cluster\_count} \times \mathrm{feature\_count}]` table with the
```
andreynv: we need to say or refer to the description of how to produce those initial centroids
need one style: table of k initial centroids of dimension p, or of k initial p-dimensional centroids

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
+   initial centroids, where each row stores one centroid.
```
andreynv: also, if we specify expected layout, do we need to add/explain pre-conditions somehow?

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
+      int64_t get_iteration_count() const;
       double get_objective_function_value() const;
    };
 
-   } // namespace onedal::kmeans
+model ``kmeans::model`` ``[read-only]``
+   The trained K-means model.
 
+labels ``table`` ``[read-only]``
+   :math:`[\mathrm{observation\_count} \times 1]` table with assignments of cluster
```
andreynv: table size is required
labels vs cluster indices - minor inconsistency. Need to introduce and fix the terminology across the section

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
 
-.. code-block:: c++
+iteration_count ``int64_t`` ``[read-only]``
+   Actual number of iterations performed by the algorithm.
 
-   int main() {
-      auto kmeans_desc = onedal::kmeans::desc{}
-         .set_cluster_count(10);
+   Invariants
+      | ``iteration_count > 0``
```
andreynv: can we stop immediately in the corner case when by initial centroids are the real centroids?

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
+
+objective_function_value ``double`` ``[read-only]``
+   The value of the objective function.
+
+   Invariantss
```
andreynv: misprint

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -132,22 +222,13 @@ Input
       infer_input& set_data(const table&);
```
andreynv: what would be the logic for the use of method argument names?

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
+iteration_count ``int64_t`` ``[read-only]``
+   Actual number of iterations performed by the algorithm.
 
-   int main() {
-      auto kmeans_desc = onedal::kmeans::desc{}
-         .set_cluster_count(10);
+   Invariants
+      | ``iteration_count > 0``
+
+objective_function_value ``double`` ``[read-only]``
```
andreynv: we return it as a double, what is probably good. The question would be when we use table of size 1x1 for parameters/results/etc and when we switch to scalars

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
 ~~~~~~
-.. code-block:: c++
-
-   namespace onedal::kmeans {
+::
 
    class train_result {
    public:
       const model& get_model() const;
       const table& get_labels() const;
```
andreynv: is this possible that in some cases the use will not need labels for training data at all?

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
+objective_function_value ``double`` ``[read-only]``
+   The value of the objective function.
+
+   Invariantss
+      | ``objective_function_value >= 0.0``
 
-      onedal::train(kmeans_desc, x_train, x_train_centroids);
-   }
 
+Operation semantics
```
andreynv: it seems we need a better structuring to differentiate parameters and new subsections

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
+::
 
-Inference
----------
+   kmeans::train_result train(const kmeans::desc&, const kmeans::train_input&);
+
+**Preconditions**
+   | ``input.data.is_empty == false``
+   | ``input.initial_centroids.is_empty == false``
+   | ``input.data.column_count == input.initial_centroids.column_count``
```
andreynv: should we check somewhere that number of initial clusters is cluster_count or similar? etc

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
+      return result.get_model();
+   }
+
+   onedal::table run_inference(const onedal::kmeans::model& model,
+                               const onedal::table& new_data) {
+
+      const auto kmeans_desc = onedal::kmeans::desc<float>{}
+         .set_cluster_count(model.get_cluster_count());
+
+      const auto result = onedal::infer(kmeans_desc, model, new_data);
```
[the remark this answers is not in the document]
rlnx: desc is first, model is second, data is third

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
-      desc& set_gamma(double);
+      desc& set_cluster_count(int64_t);
+      desc& set_max_iteration_count(int64_t);
       desc& set_accuracy_threshold(double);
-      desc& set_cluster_count(std::int64_t);
-      desc& set_max_iteration_count(std::int64_t);
    };
 
-   } // namespace onedal::kmeans
+cluster_count ``int64_t`` ``default = 2``
```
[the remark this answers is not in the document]
rlnx: To me they (DBSCAN and TBB) seem very different. Ekaterina's proposal does not cover types and default values, it's about terminology, but not about API. TBB just documents each function in a class. In our case, document `set_something`/`get_something` as two different functions is redundant/useless for user. In addition they have this in the table that needs to be scrolled horizontally to read method doc to the end, which is inconvenient for reader.

@andreynv, I didn't understand this comment. Do you suggest to stick to Ekaterina's proposal or TBB's? In general, we aligned with them in terms of RST instruments that was used, some details diverge.

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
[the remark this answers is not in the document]
rlnx: This will be part of coding guidelines, TBB come with the same question. `PascalCase` for template parameters. Parameter order is the same as in DAAL.

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: We does not specify layout in terms of homogen, soa, etc. We can still store one centroid per row in any of the mentioned layouts. It says that row `i` contains centroid for cluster `i`. Underneath we can use arbitrary layout. Concept of row/column is not tighten to layout, it's part of the table definition. 

Well, we can remove this information, but how user is supposed to interact with the centroids? How can I extract centroid for the cluster `i`? How to use it, if in the newer version we could change behavior, e.g., return transposed matrix, where we store centroids by columns. 

From the other hand, if we do not specify layout in spec, the other oneDAL implementation would store them, for example, in column-major layout (each column stores one centroid), that means our oneDAL impl 1 and oneDAL impl 2 would be incompatible.

If we do not want to say anything about centroids layout, **I would suggest to not provide API based on table at all.**

rlnx: > table with k centroids of dimension p each?

Do you suggest to use `k` and `p` instead `cluster_count` and `feature_count`.

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: `centroids` is an attribute of the model, defined earlier. Do you think it is not clear for reader?

Annotation 2020-03-21 183646

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: Yes, it is a plan

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: Yes, there is use case. In DAAL, we have `assignFlag` parameter to control that. For this version of spec, I decide to not include it, because it looks inconsistent with `ResultToCompute` enums in the other algs in DAAL.

We need to work out consistent approach for passing `ResultToCompute` to each alg in new API, after that I'll add this to spec and clarify invariants/postconditions. For now, I assume that labels are always computed.

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: Good question, need to work on rule for that, I do not have the answer so far.

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: The classical k-means stop condition compares objective function value at `(i-i)`-th iteration with the value at `i`-th iteration, so we need to perform at least one iteration to stop.

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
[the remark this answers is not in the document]
outoftardis: by the way, scrolling is easy to fix (it shouldn't be there in the first place), and I think we can ask Robert to update css to get rid of it

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,3 +1,5 @@
+.. highlight:: cpp
+
 =======
 K-Means
```
andreynv: after thinking offline on the level of details : it probably makes sense to to have a detailed description of the algorithmic flow that would help quickly correlate api specifics (distance, accuracy threshold, number of iterations, objective function, convergence etc) with this flow. We can re-use the info from the Programming Guide. However, it does not contain this info for k-means
Also, what we are going to do with the processing of empty clusters? Do we enable the specific technique in oneDAL k-means?

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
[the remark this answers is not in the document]
andreynv: https://rlnx.github.io/oneapi-spec/elements/oneDAL/source/algorithms/clustering/kmeans.html, principally follows the same style. The difference however in coloring/other details. DAL, VPL, TBB etc do this formatting differently. The right style/formatting should be discussed/agreed with Robert. Another goal-opportunity to easily refer  from different section of the spec to the specific method of k-means or its individual parameter

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
andreynv: ok for now

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
andreynv: yes, it works

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
andreynv: My comment was different, about internal storage of centroids: we can use any format we prefer; the public method however returns it in pre-defined layout, which obviously requires the  description.  If tomorrow we want to return them in column format we will add respective method and properly  describe it/resulting layout

Re k/p topic: the desire is to have the notation uniformly used across k-means section. One possible option to introduce, e.g., k and then immediately note cluster_count it in intro section ("we would note number of clusters k as cluster count below"). Another option is when mentioning feature_count, correlate it to k Need to decide what option would work

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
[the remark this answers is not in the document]
andreynv: Another observation about documenting methods
we have 
set/get for cluster_count 
but describe it as 
cluster_count int64_tdefault = 2 (std::int64?)
On the formal side, such documenting convention should pre-defined somewhere; agian, this style can look differently from the other oneapi components.  E.g., we do not describe constructors while the other do it.
The desire is to have the description that would clearly explain to the implementer, e.g, different constructors in say train_input: what if I do not pass initial centroids, what would happen/why we need a dedicated constuctor, etc

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -18,109 +20,197 @@ minimize the objective function
...
+         .set_cluster_count(model.get_cluster_count());
+
+      const auto result = onedal::infer(kmeans_desc, model, new_data);
+
+      print_table("labels", result.get_labels());
+   }
+
+
 ---
 API
```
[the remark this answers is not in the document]
andreynv: I asked Nikolay to add

outoftardis:
I'd suggest adding some sort of TOC somewhere at the top. What you have right now at the [Clustering page](https://rlnx.github.io/oneapi-spec/elements/oneDAL/source/algorithms/clustering/index.html) would be useful to have in K-means as well. Right now this page is not easy to navigate through, and the more content you add, the harder it is to navigate.

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: Added Lloyd's algorithm description, will add clarification about empty clusters in the future

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: I've updated style of class description, now it's easier to differentiate parameters and other things

rlnx: May be we need to work on better solution after 0.7

source/elements/oneDAL/source/bibliography.rst
```diff
@@ -4,8 +4,18 @@
...
+   Version:2.1 Document Revision:24
+   Available from `opencl-2.1.pdf <https://www.khronos.org/registry/OpenCL/specs/opencl-2.1.pdf>`_
+
+.. [SYCLSpec]
+   Khronos®OpenCL™ Working Group — SYCL™ subgroup, SYCL™ Specification
+   SYCL™ integrates OpenCL™ devices with modern C++, Version 1.2.1
+   Available from `sycl-1.2.1.pdf <https://www.khronos.org/registry/SYCL/specs/sycl-1.2.1.pdf>`_
 
 .. [Lloyd82]
    Stuart P Lloyd. *Least squares quantization in PCM*. IEEE Transactions on
```
andreynv: @outoftardis, this is for our records, or you probably have the immediate answer. We do bibliography in this style, and I assume this is okay.  However how do the other components do their bibliography?

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: Added description of stop criteria

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: Currently there is no `kmeans_init` alg in the spec, will add reference once I add it

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+
 =======
 K-Means
 =======
 
-K-Means is a clustering algorithm that partitions :math:`n` observations into
-:math:`k` clusters minimizing some criterion within each cluster. Each cluster
-is characterized by a representative point, called *a centroid*, and a cluster
-radius.
+The K-Means algorithm partitions :math:`n` samples into :math:`k` *clusters*
```
andreynv: term 'samples' is used, and in terminology section we seem to introduce different

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+Given the training set :math:`X = \{ x_1, \ldots, x_n \}` of
+:math:`p`-dimensional samples and a positive integer :math:`k`, the problem is
+to find a set :math:`C = \{ c_1, \ldots, c_k \}` of :math:`p`-dimensional
+centroids that minimize the objective function
+
+.. math::
+   \Phi_{X}(C) = \sum_{i = 1}^n d^2(x_i, C),
+
+where :math:`d^2(x_i, C)` is the squared Euclidean distance from :math:`x_i`
+to the closest center in :math:`C`,
```
andreynv: centroid?

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+
+.. math::
+   d^2(x_i, C) = \min_{1 \leq j \leq k} \| x_i - c_j \|^2, \quad 1 \leq i \leq k.
+
+Expression :math:`\|\cdot\|` denotes :math:`L_2` norm.
```
andreynv: L_2 - another candidate to the table with the lib math notations

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+centroids that minimize the objective function
+
+.. math::
+   \Phi_{X}(C) = \sum_{i = 1}^n d^2(x_i, C),
+
+where :math:`d^2(x_i, C)` is the squared Euclidean distance from :math:`x_i`
+to the closest center in :math:`C`,
+
+.. math::
+   d^2(x_i, C) = \min_{1 \leq j \leq k} \| x_i - c_j \|^2, \quad 1 \leq i \leq k.
```
andreynv: i<=n, not k?

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+denotes a number of the current iteration, e.g., :math:`C^{(t)} = \{ c_1^{(t)},
+\ldots, c_k^{(t)} \}` is the set of centroids at the :math:`t`-th iteration. The
+method requires the initial centroids :math:`C^{(1)}` to be specified.
+
+**(1) Assignment step:** Each sample :math:`x_i` is assigned to the nearest centroid.
+
+.. math::
+   y_i^{(t)} = \mathrm{arg}\min_{1 \leq j \leq k} \| x_i - c_j^{(t)} \|^2, \quad 1 \leq i \leq n.
+
+Each observation from the training set :math:`X` is assigned to exactly one
```
andreynv: here we use 'observation'

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+
+   :tparam Float: The floating-point type that the algorithm uses for
+                  intermediate computations. Can be :expr:`float` or :expr:`double`.
+
+   :tparam Method: Tag-type that specifies variant of K-Means algorithm. Can be
+                   :expr:`method::lloyd` or :expr:`method::by_default`.
+
+   .. function:: desc()
+
+      Creates new instance of descriptor with the default attribute values.
```
andreynv: 'Create**s** is aligned with TBB style of description, VPL seems to do 'create'
TBB uses 'constructs'. Looks like we will need simple guidelines or a decision each library can do it differently

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+
       const table& get_centroids() const;
+      int64_t get_cluster_count() const;
    };
 
+.. class:: model
+
+   .. function:: model()
+
+      Creates model with default attribute values.
```
andreynv: what is 'attribute' in this context?

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
-~~~~~~~~~~~~~
+      Creates result of the training operation with the default attribute
+      values.
 
-.. code-block:: c++
 
-   int main() {
-      auto kmeans_desc = onedal::kmeans::desc{}
-         .set_cluster_count(10);
+   .. member:: kmeans::model model = kmeans::model()
```
andreynv: this line seems to require space between 'models'

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+
+      :math:`n \times p` table with the data to be assigned to the clusters,
+      where each row stores one sample.
+
+      Getter & Setter
+         | ``const table& get_data() const``
+         | ``infer_input& set_data(const table&)``
+
+
+   .. member:: kmeans::model model = kmeans::model()
```
andreynv: Is space required?

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+      values.
+
+
+   .. function:: infer_input(const kmeans::model& model)
+
+      Creates input for the inference operation with the given :expr:`model`, the
+      other attributes get default values.
+
+
+   .. function:: infer_input(const kmeans::model& model, const table& data)
```
andreynv: generally, later need double-check style, spaces, etc

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
andreynv: ok

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
outoftardis: a related q of style: in the output there is no visible difference between functions and members, but a description of a function begins with the verb (which makes sense), while the members describe entities (also makes sense). This confuses me as a reader. @rscohn2  I'd suggest updating css to differentiate functions and members.

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
-Given the set :math:`X = \{ x_1, \ldots, x_n \}` of :math:`p`-dimensional
-feature vectors and a positive integer :math:`k`, the problem is to find a set
-:math:`C = \{ c_1, \ldots, c_k \}` of :math:`p`-dimensional vectors that
-minimize the objective function
+--------------
+Lloyd's method
+--------------
+
+The Lloyd's method [Lloyd82]_ consists in iterative updates of centroids by
+applying the alternating *Assignment* and *Update* steps, where :math:`t`
```
outoftardis: I'd make sections for those steps, and add here a link to them instead of italics.

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+denotes a number of the current iteration, e.g., :math:`C^{(t)} = \{ c_1^{(t)},
```
outoftardis: "the index" instead of "a number" seems more accurate

and "e.g." seems irrelevant here, you can put ":" there instead of it, or split the sentence in two

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+\ldots, c_k^{(t)} \}` is the set of centroids at the :math:`t`-th iteration. The
+method requires the initial centroids :math:`C^{(1)}` to be specified.
+
+**(1) Assignment step:** Each sample :math:`x_i` is assigned to the nearest centroid.
+
+.. math::
+   y_i^{(t)} = \mathrm{arg}\min_{1 \leq j \leq k} \| x_i - c_j^{(t)} \|^2, \quad 1 \leq i \leq n.
+
+Each observation from the training set :math:`X` is assigned to exactly one
+centroid such that :math:`X` is partitioned to :math:`k` disjoint sets
```
outoftardis: change "such that X is partitioned to k ..." into one of these:
* "so that X is partitioned into k..."
* "in order to split X into k ..."

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+Lloyd's method
+--------------
+
+The Lloyd's method [Lloyd82]_ consists in iterative updates of centroids by
+applying the alternating *Assignment* and *Update* steps, where :math:`t`
+denotes a number of the current iteration, e.g., :math:`C^{(t)} = \{ c_1^{(t)},
+\ldots, c_k^{(t)} \}` is the set of centroids at the :math:`t`-th iteration. The
+method requires the initial centroids :math:`C^{(1)}` to be specified.
+
+**(1) Assignment step:** Each sample :math:`x_i` is assigned to the nearest centroid.
```
outoftardis: I'd use active voice here as it's done in the next step " Assign each sample ..."

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
-   \Phi_{X}(C)=\sum_{x_i \in X}d(x_i, C),
+   S_j^{(t)} = \big\{ \; x_i \in X : \; y_i^{(t)} = j \; \big\}, \quad 1 \leq j \leq k.
+
+**(2) Update step:** Recalculate centroids by averaging samples assigned to each
+cluster.
+
+.. math::
+   c_j^{(t + 1)} = \frac{1}{|S_j^{(t)}|} \sum_{x \in S_j^{(t)}} x, \quad 1 \leq j \leq k.
+
+The steps (1) and (2) are performed until the following **stop condition**,
```
outoftardis: Active voice is better: "Perform steps ..."

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+
+.. math::
+   \sum_{j=1}^k \big\| c_j^{(t)} - c_j^{(t+1)} \big\|^2 < \varepsilon,
+
+is satisfied or number of iterations exceeds the maximal value :math:`T` defined
+by the user.
```
outoftardis: I'd change it into:

```
Perform steps (1) and (2) until one of the following happens:
- The stop condition is satisfied:
  <equation here>
- The number of iterations exceeds the pre-defined maximal value T
```

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
 
+.. namespace:: onedal::kmeans::method
+.. struct:: lloyd
+
+   Tag-type that denotes `Lloyd's method`_.
+
+
+.. type:: by_default = lloyd
+
+   Alias tag-type for the Lloyd's method.
```
outoftardis: also add a reference here to Lloyd's method

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
-   } // namespace onedal::kmeans
+.. namespace:: onedal::kmeans
+.. class:: template<typename Float = float, \
+                    typename Method = method::by_default> \
+           desc
+
+   :tparam Float: The floating-point type that the algorithm uses for
+                  intermediate computations. Can be :expr:`float` or :expr:`double`.
+
+   :tparam Method: Tag-type that specifies variant of K-Means algorithm. Can be
```
outoftardis: not variant, but "an implementation" maybe?

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
outoftardis: "a new instance of the descriptor"

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
outoftardis: a model with the default

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+      The trained K-means model.
+
+      Getter
+         | ``const model& get_model() const``
+
+
+   .. member:: table labels = table()
+
+      :math:`n \times 1` table with assignments of cluster indices to samples in
+      the input data.
```
outoftardis: table with the cluster indices assigned to the samples in the input data

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+
+      Getter
+         | ``const table& get_labels() const``
+
+
+   .. member:: std::int64_t iteration_count = 0
+
+      Actual number of iterations performed by the algorithm.
```
outoftardis: I'd remove actual
"The number of ..."

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,153 +1,494 @@
...
+
+      The trained K-means model.
+
+      Getter
+         | ``const model& get_model() const``
+
+
+   .. member:: table labels = table()
+
+      :math:`n \times 1` table with assignments of cluster indices to samples in
```
outoftardis: is it samples or observations?

source/elements/oneDAL/source/bibliography.rst
outoftardis: I couldn't find bibliography in other spec components (except for [oneMKL](https://spec.oneapi.com/versions/0.6.0/oneMKL/GUID-732F9EE1-BCEC-4D9B-9B93-AF5499B21140.html), which follows the same style as oneDAL)

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: I'd like to change terminology, observation/feature vector -> sample

rlnx: I'l update corresponding sections

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: Looks like a problem with CSS.

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: Looks like a problem with CSS.

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: Attribute aggregates setter and setter functions to a single entity called _attribute_, we'll need to introduce this terminology somewhere in common section

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
rlnx: It's HTML generator problem, RST is ok

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
```diff
@@ -1,25 +1,70 @@
...
+.. note::
+   In the general case, :math:`d` may be an arbitrary distance function. Current
+   version of the oneDAL spec defines only Euclidean distance case.
+
+
+--------------
+Lloyd's method
+--------------
+
+The Lloyd's method [Lloyd82]_ consists in iterative updates of centroids by
```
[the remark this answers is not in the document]
rlnx: I mean "consists in", as far as I know this is widely used in math/physics "method consists in iterative solving". If I change to "consists of" the meaning will be different.

https://english.stackexchange.com/questions/61600/consist-in-vs-consist-of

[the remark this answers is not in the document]
outoftardis: ok, agree

source/elements/oneDAL/source/algorithms/clustering/kmeans.rst
andreynv: looking from the perspective of algorithms and tables with features/observations in 0.7, I'd decide on it a bit later

rlnx: ok

andreynv:
minor:  **int64**_t get_cluster_count() const; can be done after 0.7, among other things

Issue #86, opened 2020-03-23
Update css


outoftardis:
Long lines of text do not wrap properly in a table, and the reader has to scroll to see content:
- [example №1](https://spec.oneapi.com/versions/latest/elements/oneTBB/source/algorithms/splittable_concept/proportional_split_cls.html#members)
- [example №2](https://spec.oneapi.com/versions/latest/elements/oneTBB/source/general/notation.html)

Related:

There are 2 css files with identical content:
- [in `/source/_static`](https://github.com/oneapi-src/oneapi-spec/blob/master/source/_static/custom.css)
- [in `/source/elements/oneTBB/source/_static`](https://github.com/oneapi-src/oneapi-spec/blob/master/source/elements/oneTBB/source/_static/custom.css)

It seems that the second one should be removed.

rscohn2:
@outoftardis  Do you know how to update the css to wrap instead of scroll?

We can build the full doc or TBB standalone. I don't know how to share one directory for both. That is why it is duplicated.

outoftardis:
> Do you know how to update the css to wrap instead of scroll?

Read the Docs's [instructions](https://rackerlabs.github.io/docs-rackspace/tools/rtd-tables.html)

outoftardis:
> We can build the full doc or TBB standalone. I don't know how to share one directory for both. That is why it is duplicated.

I'd suggest to introduce a script into the building process that copies the main css file into the right directory each time the tbb is built as a standalone. Maybe at this point having two identical css files seems ok, but you are going to introduce changes to css over time, and this is additional work to have these changes implemented in two places each time. And what if other libraries would need this "build as a standalone" feature?

rscohn2:
Having the script sync it should work well, and I can add the css at the same time. Thanks.

rscohn2:
Adding @rsdubtso

rscohn2:
The tables look ok now https://oneapi-src.github.io/oneAPI-spec/elements/oneTBB/source/notation.html

PR #99, opened 2020-03-26
add DAL, make level zero first

rscohn2:
@outoftardis: Adding DAL. Formatting is not good. sphinx version does not look the same as github and has these problems:

* I don't like the horizontal scrollbar.
* First column should be visually different because it is the container for the other specs. 
* blank line under first row, ...

Suggestions?


outoftardis:
* The scrollbar is what I mentioned in #86 
* blank line just shows it's a header, I think (I have it in other docs as well -- it's probably a ReadTheDocs thing)

rscohn2:
I will give the CSS thing a try. I was thinking it makes text wrap instead of scroll. If it expands the column, then that would fix my problem

outoftardis:
[First column looking different in RTD](https://sphinx-rtd-theme.readthedocs.io/en/stable/demo/lists_tables.html#list-tables)

You'd need to use `list-table` directive for a table with `:stub-columns: 1`

outoftardis:
> I will give the CSS thing a try. I was thinking it makes text wrap instead of scroll. If it expands the column, then that would fix my problem

It does wrap text, so I'm wrong here, it won't help with so many columns. This might help though:

```
/* make the page width fill the window */
.wy-nav-content {
   max-width: none;
}
```

rscohn2:
If I set max-width to none in inspector, it makes the column wider. Putting the code in custom.css doesn't seem to do that.

outoftardis:
try referencing `custom.css` from `_templates\layout.html`:

```
<link href="{{ pathto("_static/custom.css", True) }}" rel="stylesheet" type="text/css">
```

[(c) stack overflow](https://stackoverflow.com/a/33626140/12767846)

rscohn2:
I tried all the suggestions in the stackoverflow article and none of them work for me. I will set aside formatting for now.

PR #132, opened 2020-04-29
style guide

rscohn2:
@oneapi-src/oneapi-spec-maintainers @Jianhi-Li

I reformatted Andrey's document and did some editing.

It may not match exactly what Andrey said.

### DIFF
--- a/style/content.rst
+++ b/style/content.rst
@@ -11,7 +11,7 @@ Principles
 
 * Strive for consistency across components
 
-  * Newer specs expected to following pre-existing practices
+  * Especially newer specs
   * Component must be consistent with itself
 
 * Target audience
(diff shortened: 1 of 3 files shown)


style/content.rst
```diff
@@ -0,0 +1,143 @@
...
+* Spec must be publishable as part of unified doc and as standalone
+  describing a single component
+
+  * when published as a standalone, it will contain some sections that
+    are normally at top-level. e.g. bibliography
+  
+Structure of Document
+=====================
+
+* Top-level spec contains sections for information common to all
```
rsdubtso, line 34: I would prefer a more traditional way to word it: 'should contain', etc. The rest of the document lacks verbs in many places. @andreynv : can you please edit the document to make it more readable? For example I'm not sure I understand what the 'Conventions' bullet really lists.

andreynv: As mentioned on the meeting, my doc - quick summary/theses on the spec structure considerations to discuss on the bridge, w/o any goal to have 'shall do'. The published version - reformatted one that can deviate from mine. It makes sense first to have thoughts of @rscohn2 on the next steps.

style/content.rst
```diff
@@ -0,0 +1,143 @@
...
+* Component-level sections for information specific to the component
+  
+  * Introduction: defines the scope of the SW
+  * Glossary: defines terms used to describe specs
+  * Math notations
+  * Namespace-related information: e.g., the first level namespace,
+    the lib introduces additional namespaces based on some internal
+    logic
+  * header files
+  * SW API usage with examples demonstrating the API of the lib
```
napetrov, line 59: There will be at least 1 API for "get_spec_version" that should be common for all libs i think.   So user will have same way for getting it regardless of product.   

We ether can follow approach as we have with other design decisions there we align on API/design but have own implementations.  Or move such things to common place. 

The main difference is that for spec version we should provide same API and same behavior across libs.

rscohn2: I will add some requirement to the common section, but the actual API needs to be specified in each component.

andreynv: 'Conventions' would contain the list of the cross-library consistency topics incl spec version and version format. Respective proposal on the spec version suggests have it in the specs but gives the flexibility how to implement it. Whether to have the lib version in specs or not - up to the library

style/content.rst
rsdubtso: The problem is that having missed the meeting I don't really understand how to read this. For example, what does 'Basic scalar data types' mean under the 'Conventions' ? Does it mean that my spec needs such section?

andreynv: The idea - have 'xIPL' conventions in the single place, on the top level. The components would have a short note on compliance with the conventions, alternatively, the note only in case of specific deviations

PR #133, opened 2020-05-09
Update oneDNN spec to beta6 content

rsdubtso:


rscohn2:
It looks good. I will let you merge because it looks like you force pushed another update after I looked at the diff.

rsdubtso:
> It looks good. I will let you merge because it looks like you force pushed another update after I looked at the diff.

Thanks. I've force-pushed the table-formatting change. Will merge as soon as everything is green.

PR #134, opened 2020-05-11
[oneDAL] Glossary section

michael-smirnov:
- Terminology renamed to glossary.
- All terms divided onto three groups: ML, oneDAL specific, and common. In my opinion, common terms should be moved to the top-level as proposed in #122.
- Some terms inside the ML group and oneDAL group are reworked.

The compiled version of glossary: https://michael-smirnov.github.io/oneapi-spec/elements/oneDAL/source/glossary.html


michael-smirnov:
@andreynv @rlnx @outoftardis what is your opinion about dividing the glossary onto several parts (ML/oneDAL specific)? 
I see that some of the terms can be present in both sections and, for example, in oneDAL specific section may have more detailed or slightly different meanings (e.g. *model*, *parameter*, *algorithm*, ...).

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+
+    Kernel
+        Code written in OpenCL [OpenCLSpec]_ or :term:`SYCL` and executed on a GPU device.
+
+    SPIR-V
+        Standard Portable Intermediate Representation - V is a language for intermediate representation of compute kernels.
+
+    SYCL
+        SYCL(TM) [SYCLSpec]_ --- high-level programming model for OpenCL(TM) that enables code for heterogeneous
+        processors to be written in a "single-source" style using completely standard C++.
```
rlnx, line 114: Please take a look at #122, those terms are going to be common for the entire spec

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+        A value of a single :term:`label`. For example, response may be a flag
+        indicating that an observation belongs to a particular class.
+
+    Sample
+        A synonym for :term:`instance`.
+
+    Training
+        A search process of the best parameter values of the :term:`model`.
+
+    Variable
```
rlnx, line 63: Alias for the attribute/feature

rlnx:
> @andreynv @rlnx @outoftardis what is your opinion about dividing the glossary onto several parts (ML/oneDAL specific)?

> some of the terms can be present in both sections

The last proposition sounds misleading to me as the term is supposed to have different meaning depending on a context. Probably it's not we want introducing glossary. In a perfect world, if one sees the term in the text, its meaning should be clear without understanding the rest of the writing.


> I made it mostly for simplicity (not to mix terms from different areas)

Having ML section helps understand the reader the language we have chosen for ML as there a lot of different terminology/points of view in this area. Given the fact @rscohn2 proposed to sort a glossary alphabetically, it also helps to group the terms logically, e.g., it's strange to see `Classification` next to `Batch mode`.

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+
+.. glossary::
+    :sorted:
+
+    Batch Mode
+        The computation mode for an algorithm in |dal_short_name|, where all the data needed for
+        computation is available at the start and fits the memory of the device on which the computations are performed.
+
+    Immutability
+        The object is immutable when it is not possible to change its state after it is constructed.
```
outoftardis, line 77: I'd change to: 
The object is immutable if it is not possible to change its state after creation.

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+.. glossary::
+    :sorted:
+
+    Attribute
+        A particular property or quality of a real object or an event.
+        Has a defined type and domain.
+
+    Classification
+        A supervised machine-learning problem to assign categorical :term:`labels<Label>`
+        to :term:`feature vectors<Feature vector>` describing input objects.
```
outoftardis, line 19: * problem of assigning
* do you need "describing input objects" here? don't feature vectors always describe input objects?

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+
+    Feature vector
+        A vector that encodes information about single real object or an event.
+        Contains at least one :term:`feature`.
+
+    Inference
+        A process of obtaining outputs from the :term:`model` created on the :term:`training` stage
+
+    Instance
+        A vector of at least one :term:`feature` value and zero or more :term:`label<Label>` values.
```
outoftardis, line 35: what about regression? you don't always have labels

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+        Each of :term:`labels<Label>` and :term:`features<Feature>` can encode different objects or events.
+
+    Label
+        :term:`Feature` of some object or event which is present on the :term:`training` stage but
```
outoftardis, line 39: **at** the training stage

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+        unavaliable on the :term:`inference` stage.
```
outoftardis, line 40: * it's not only unavailable, but it's something you predict at the inference stage -- I think it should be mentioned
* **at** the inference stage

outoftardis:
I think that in ML section it makes sense to mention synonyms. Not give them a separate entry in the glossary, but give a definition to the term we are using and then mention that across ML field you may encounter other synonyms of that term, such as <...> .

outoftardis:
> @andreynv @rlnx @outoftardis what is your opinion about dividing the glossary onto several parts (ML/oneDAL specific)?
> I made it mostly for simplicity (not to mix terms from different areas), but I also see that some of the terms can be present in both sections and, for example, in oneDAL specific section may have more detailed or slightly different meanings (e.g. model, parameter, algorithm, ...).

Dividing into sections is a good idea, but having the same terms appear in more than one section is not. If you have a definition for model in ML, but you also want to specify how model is understood in oneDAL specs, you do it under the same entry. It's still a ML term.

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+Glossary
+=========
+
+Machine learning terms
+======================
+
+.. glossary::
+    :sorted:
+
+    Attribute
```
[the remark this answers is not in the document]
andreynv: We might have a 'main' term and then list its synonyms in the same line as Ekaterina suggested somewhere else

source/elements/oneDAL/source/glossary.rst
andreynv, line 40: When I train a model I can use the part of the available data set to do that. The rest of the data set incl labels can be used to test the model by predicting and comparing vs known label

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+
+    Instance
+        A vector of at least one :term:`feature` value and zero or more :term:`label<Label>` values.
+        Each of :term:`labels<Label>` and :term:`features<Feature>` can encode different objects or events.
+
+    Label
+        :term:`Feature` of some object or event which is present on the :term:`training` stage but
+        unavaliable on the :term:`inference` stage.
+
+    Model
```
andreynv, line 42: general question: do we rely on some literature that introduces the terms or define our self

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+    Online Mode
+        The computation mode for an algorithm in |dal_short_name|, where the data needed for computation
+        becomes available in parts over time.
+
+    Workload
+        A task of applying a |dal_short_name| algorithm to a dataset.
+
+Other
+=====
+TODO: this section looks like a general one for the whole oneAPI spec.
```
andreynv, line 88: yes, it should be in the common glossary

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+        computation is available at the start and fits the memory of the device on which the computations are performed.
+
+    Immutability
+        The object is immutable when it is not possible to change its state after it is constructed.
+
+    Online Mode
+        The computation mode for an algorithm in |dal_short_name|, where the data needed for computation
+        becomes available in parts over time.
+
+    Workload
```
andreynv, line 83: in addition to those terms, we would need terms related to features and necessary to describe data management: continuous feature, data layout, ....

andreynv:
@andreynv @rlnx @outoftardis what is your opinion about dividing the glossary onto several parts (ML/oneDAL specific)?

We would have a common glossary in a separate place.
On ML/oneDAL specific and duplication - when we add a new term how to decide where to add it? A logical grouping potentially makes sense

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+        :term:`inference` on a new :term:`dataset`.
+
+    Observation
+        A synonym for :term:`instance`.
+
+    Prediction
+        - A synonym for :term:`inference`.
+        - An output of the :term:`trained<Training>` :term:`model` when provided with an input :term:`instance`.
+
+    Response
```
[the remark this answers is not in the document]
michael-smirnov: Cannot see why. Labels are for both discrete and continuous features. Are you asking to delete an example from this term?

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+        A value of a single :term:`label`. For example, response may be a flag
+        indicating that an observation belongs to a particular class.
+
+    Sample
+        A synonym for :term:`instance`.
+
+    Training
```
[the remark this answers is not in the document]
michael-smirnov: Changed to "A search process of the best model" - is it more accurate?

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Please look at the description of this PR. I mentioned that this section should be deleted once it moved to the top level.

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,117 @@
...
+
+    Feature
+        A pair of :term:`attribute` information and a corresponding value.
+
+    Feature vector
+        A vector that encodes information about single real object or an event.
+        Contains at least one :term:`feature`.
+
+    Inference
+        A process of obtaining outputs from the :term:`model` created on the :term:`training` stage
```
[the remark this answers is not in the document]
michael-smirnov: Reformulated.

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Labels are applicable for the regression task too. For unsupervised learning tasks, there are no labels present, that's why I wrote "zero or more".

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Reformulated.

michael-smirnov: Yes, but you are talking about test or validation stages, not inference. Do you mean I should mention these stages too?

source/elements/oneDAL/source/glossary.rst
michael-smirnov: I formulate it by myself, but with accordance to external resources. I use [this](https://ml-cheatsheet.readthedocs.io/en/latest/glossary.html) and [this](https://developers.google.com/machine-learning/glossary) links and Peter Flach's book to check the terms.

source/elements/oneDAL/source/glossary.rst
michael-smirnov: I mean, this section should be deleted when common glossary appears. Now I don't want to remove it because we have a references on the spec to these terms, which will be broken after removal. @rlnx @andreynv @outoftardis what is your opinion, should I delete it now?

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Yes, I've added some of the terms. Now, there is a question about *data layout* - should we introduce this term and entity or not. I think we need to discuss metadata API in details, I will prepare for it. Now I don't introduce this term.

source/elements/oneDAL/source/glossary.rst
andreynv: I mean that we formally use the inference stage to validate the model. I assume you are talking about inference stage where you use earlier unseen data. Probably 1) a better wording is required to describe what the 'label' is, 2) unavailable label is more about specifics/def of supervised learning, and that place would be better to discuss it

source/elements/oneDAL/source/glossary.rst
andreynv: I'd have it here until a common mechanism for dictionaries/bibliographies is agreed and implemented on oneapi level

source/elements/oneDAL/source/glossary.rst
andreynv: The specifics here is that while we can use 'data layout', 'data format' etc in the specific dal sense, the real sense of the terms is generally different. The idea-to make sure we have defs aligned with community's understanding to avoid confusion; moreover it would be used in apis somewhere

source/elements/oneDAL/source/glossary.rst
andreynv: option to think: if there is a well recognized a book/source with thedefs we might rely on it and then refer that we use this source as the base for our ml glossary in the volumes necessary for the lib

source/elements/oneDAL/source/glossary.rst
andreynv: do we use 'label' in the context of regression?

source/elements/oneDAL/source/glossary.rst
outoftardis: I agree with @andreynv. But don't call this section "Other".  oneAPI terms, maybe? or common oneAPI terms?

source/elements/oneDAL/source/glossary.rst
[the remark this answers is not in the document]
outoftardis: It's grammatically incorrect (search **for**)
But I also don't like this definition because it doesn't say anything about how you determine which model is the best. What exactly are you searching for? What do you have at the beginning of that search (e.g. you need a training set and a target variable).
I'd mention parametric and non-parametric methods here (q: should there be entries for them in the glossary?)

source/elements/oneDAL/source/glossary.rst
outoftardis: label in the context of regression sounds weird to me, mostly because in my understanding labels are categorical

source/elements/oneDAL/source/glossary.rst
outoftardis: why are we describing what label is as "something that present at the stage A and unavailable at the stage B"? it is not a definition of a label.
I'm still confused how labels are present in regression, but if we are talking about labels in classification, I'd highlight that 1) labels are categorical 2) we talk about labels when we discuss the response (outcome, target variable), not the independent variables 3) it's relevant to supervised learning, as Andrey mentioned above 4) you can mention how labels are used at the training stage vs inference 5) might be a good idea to give examples

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,159 @@
...
+
+    Continuous feature
+        A :term:`feature` with values in a domain of real numbers.
+
+        Synonyms: quantitative feature.
+
+    Dataset
+        A collection of :term:`observations<Observation>`.
+
+    Feature
```
outoftardis, line 31: makes sense to link to the features of different types

and a general q: do we maybe want to list related terms? If something that might be relevant for the reader but not specifically appears in the definition

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,159 @@
...
+=========
+
+Machine learning terms
+======================
+
+.. glossary::
+    :sorted:
+
+    Categorical feature
+        A :term:`feature` with a discrete set of possible values without ordering.
```
outoftardis, line 14: I think it's finite, not discrete

rlnx:
@michael-smirnov, minor comment regarding line length, mostly for the future. It is easier to review changes, if line length is about 80--100 symbols as git compares files line by line. Even a small change in a line will be interpreted like one line is removed and the other is added.

I'd recommend to look at [Rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap).

source/elements/oneDAL/source/glossary.rst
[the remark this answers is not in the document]
rlnx: Is there any book/article that uses the term `label` for regression problems? I suggest to stick to more adopted terminology.

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,159 @@
...
+        predict :term:`responses<Response>` of the :term:`labels<Label>`.
+
+        Synonyms: prediction.
+
+    Inference set
+        An unlabeled :term:`dataset` used at the :term:`inference` stage to make an outputs from the :term:`model`.
+
+    Label
+        :term:`Feature` of some object or event which is unavailable at the :term:`inference` stage
+        and needs to be predicted.
```
rlnx, line 55: It's too debatable to use the `label` for continuous variables, [the same question here](https://github.com/oneapi-src/oneAPI-spec/pull/134/files#r424726360). Maybe introduce more liberal term like `target variable`?

michael-smirnov: @rlnx @andreynv @outoftardis guys, I am not inventing my own ML terms. Here are the sources which have this term:
- Peter Flach, Machine Learning. The Art and Science of Algorithms that Make Sence of Data, Cambridge University Press, 2012. page 50.
> We furthermore distinguish between the label space `L` and the output space `Y`. The label space is used in supervised learning to label examples. <...> The most commonly encountered machine learning scenario is where the label space coincides with the output space. That is, `Y=L` <...>. This scenario covers both classification and regression. In cases where the label space and the output space differ, this usually serves the purpose of learning a model that outputs more than just a label <...>.
- Jure Leskovec, Anand Rajaraman, Jeffrey D. Ullman, Mining of Massive Datasets, 2014. page 440
> A training set consists of a set of pairs `(x, y)` <...> where `y` is a label, the classification value for `x`. <...> The type of `y` is in principle arbitrary, but there are several common and important cases. `y` is real number. In this case, the ML problem is called regression. `y` is a boolean value <...>. In this class the problem is binary classification. <...>
- Google`s ML glossary: [label](https://developers.google.com/machine-learning/glossary#label)
- [open-source ML glossary](https://ml-cheatsheet.readthedocs.io/en/latest/glossary.html#glossary-observation)
> Label
The “answer” portion of an observation in supervised learning.
- Direct [question](https://stackoverflow.com/questions/40898019/what-is-the-difference-between-a-feature-and-a-label) about the label on stack overflow.
>Briefly, feature is input; label is output. This applies to both classification and regression problems.

If you think that the term "target variable" or some other term is better, I can change for it.

source/elements/oneDAL/source/glossary.rst
michael-smirnov: I meant (discrete data)[https://stats.stackexchange.com/questions/206/what-is-the-difference-between-discrete-data-and-continuous-data]. Maybe, it is better to reformulate it like this?
"A feature with a discrete domain without ordering between values"

source/elements/oneDAL/source/glossary.rst
michael-smirnov: I would not want to do this, because the glossary transforms into an ML guide. In my opinion, the terms should be short, like in vocabulary. The reader must be referenced to the **direct** term he wants to clarify in the spec. And this term can refer to other terms. 

@rlnx @andreynv what is your opinion?

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Yes, it might be good and probably will give a "weight" to our terminology in the reader's eyes. 
But I really don't know, can we call the resources which I list somewhere below as a "well recognized"? How to measure it? I mostly rely on the **number** of resources and their **difference** to formulate more or less "objective" definitions.
Another thing is that most of the books have very long, "continuous" terms, which need to be reformulated shortly. Can we refer to a book if we mostly reformulated the term which is described in it?
 @andreynv @rlnx can you please name some well-recognized sources which I should also consider?

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,159 @@
...
+        A :term:`feature` with a discrete set of possible values without ordering.
+
+        Synonyms: nominal feature.
+
+    Classification
+        A :term:`supervised machine-learning problem<Supervised learning>` of assigning
+        discrete :term:`labels<Label>` (:term:`categorical<Categorical feature>` or :term:`ordinal<Ordinal feature>`)
+        to :term:`feature vectors<Feature vector>`.
+
+    Continuous feature
```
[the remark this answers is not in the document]
michael-smirnov: Its debatable how to name such features. Some sources name it "numerical", some "continuous" or "quantitative". For example, Peter Flach (p.304):
> <...> we can distinguish three main kinds of feature: those with a meaningful numerical scale, those without scale but with an ordering, and those without either. We will call features of the first type quantitative <...> (another term in common use is 'continuous').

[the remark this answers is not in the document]
michael-smirnov: @rlnx @andreynv what is your opinion on how we should primarly name it? Three of my sources name it differently - Peter Flach calls it "quantitative", Jure Leskovec et all call it "numerical", Google ML glossary - "continuous". I chose the last one because it has been already used in DAAL.

source/elements/oneDAL/source/glossary.rst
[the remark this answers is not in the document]
michael-smirnov: Reformulated, please look. It is very hard to bring any details about this term because it is very common. I cannot even say that it is an optimization process, that's why I used the term "search process".

source/elements/oneDAL/source/glossary.rst
andreynv: This is the mess with terminology.
This book was suggested by the lead of ML group at TMG, http://faculty.marshall.usc.edu/gareth-james/ISL/.  It uses 'quantitative response', response in case of regression and 'binary qualitative response' for classification. I'd clearly differentiate this dependent variable for regression and classification. Response or target variable or similar for regression and label for classification. If the desire to connect them, say about response/target variable in generic sense, then specify quantitative and connect to regression with the right term, specify qualitative and connect to label

source/elements/oneDAL/source/glossary.rst
andreynv: 'discrete, continuous, ordinal, categorical' (and Mikhail A also suggested interval) were aligned with http://dmg.org/pmml/v4-1/DataDictionary.html in the lib

source/elements/oneDAL/source/glossary.rst
andreynv: the quick example from the top of my mind. During the design of dbscan the terminology that came from the paper was 'points' or so which means observation of feature vector for the lib. To make sure that the dbscan is described in properly, we introduced that synonym.  to simplify, we might have the same term and rely on it across the specs even if it would deviate from the terminology in the paper (for 'dbscan' we might refer to the paper and mention that it uses 'point' instead of 'observation')

source/elements/oneDAL/source/glossary.rst
andreynv: Mentioned one somewhere in the other comment and let you know, if I have another

source/elements/oneDAL/source/glossary.rst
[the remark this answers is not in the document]
andreynv: responded somewhere else. If we see different terms, we have freedom to stick to one of them

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,170 @@
...
+Machine learning terms
+======================
+
+.. glossary::
+    :sorted:
+
+    Categorical feature
+        A :term:`feature` with a discrete domain without ordering between values.
+
+        Synonyms: nominal feature.
```
andreynv, line 16: here we specify synonyms in one style and somewhere else in another style?

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,170 @@
...
+    Feature value
+        A single point in the domain of the :term:`feature`.
+
+    Feature vector
+        A vector that encodes information about single real object or an event.
+        Contains at least one :term:`feature value`.
+
+    Inference
+        A process of applying a :term:`trained<Training>` :term:`model` to the :term:`inference set` with a goal to
+        make outputs based on input :term:`feature vectors<Feature vector>`.
```
andreynv, line 51: make outputs may not be clean

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,170 @@
...
+        :term:`labels<Label>` is selected. :term:`Label` is one of outputs can be made by :term:`model`
+        on the :term:`inference` stage.
+
+    Model
+        A result of :term:`training` machine-learning algorithm that stores information necessary to run
+        :term:`inference` on a new :term:`dataset`.
+
+    Observation
+        A vector of at least one :term:`feature` value and zero or more :term:`label<Label>` values.
+        Each of :term:`labels<Label>` and :term:`features<Feature>` can encode different objects or events.
```
andreynv, line 70: this label should be stabilized when we close this discussion above

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,170 @@
...
+
+        Synonyms: instance, sample.
+
+    Ordinal feature
+        A :term:`feature` with a discrete set of possible values and some ordering between them.
+
+    Regression
+        A :term:`supervised machine-learning problem<Supervised learning>` of assigning
+        :term:`continuous<Continuous feature>` :term:`labels<Label>` to
+        :term:`feature vectors<Feature vector>`.
```
andreynv, line 80: the definition is still incomplete

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,170 @@
...
+
+    Response
+        A value of a single :term:`label`. For example, response may be a flag
+        indicating that an observation belongs to a particular class.
+
+    Supervised learning
+        :term:`Training` process with :term:`labeled<Label>` :term:`training set`.
+        The best :term:`model` of dependencies between input :term:`feature vector`
```
andreynv, line 88: best may sound confusing ; need a more neutral term

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,170 @@
...
+    Training
+        A process of creating a best :term:`model` (in terms of some quality criteria) based on
+        information extracted from a :term:`training set`.
+
+    Training set
+        A :term:`dataset` used at the :term:`training` stage to create a :term:`model`.
+
+    Unsupervised learning
+        :term:`Training` process where :term:`training set` does not contain information
+        of dependencies between :term:`feature vectors<Feature vector>` and outputs need
```
andreynv, line 100: information on dependencies

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,170 @@
...
+        computation is available at the start and fits the memory of the device on which the computations are performed.
+
+    Contiguous data
+        A special property of the :term:`data format`: the data are stored as one contiguous memory block.
+
+    Data format
+        The representation of internal structure of the data: this information is enough to efficiently access them.
+
+    Data type
+        An integral or floating-point type which is used to physically store the data.
```
andreynv, line 120: what term is used by c++ standard, integral, basic, scalar?

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,170 @@
...
+
+    Homogeneous data
+        A special property of the :term:`data format`: all the :term:`features<Feature>` have the same type and domain.
```
andreynv, line 123: that's interesting: both features are double, but one belongs to (0,1) and another - to (2,3). Is it homogeneous or not?

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,170 @@
...
+    Immutability
+        The object is immutable if it is not possible to change its state after creation.
+
+    Online Mode
+        The computation mode for an algorithm in |dal_short_name|, where the data needed for computation
+        becomes available in parts over time.
+
+    Reference-counted object
+        A |dal_short_name| object which stores the number of references to the unique implementation.
+        Such object has a lightweight copy constructor and assignment operator, which simply increases the number of references.
```
andreynv, line 134: it requires def of copy constructor/assignment operator - is it possible to simplify?
'simply' - remove

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Used `response` term from this book to reference output variables for both classification/regression, changed the `label` term point only to discrete features.

source/elements/oneDAL/source/glossary.rst
michael-smirnov: I specify synonyms consistently across this document. @outoftardis, is it ok to indicate synonyms as I do? Or we have some other approved/used style.

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Reformulated

source/elements/oneDAL/source/glossary.rst
michael-smirnov: - *character types` for `char`, `unsigned char`, ...
- *integer types* for `int`, `long`, `short` ...
- *integral types* for `bool` + character types + integer types

source/elements/oneDAL/source/glossary.rst
michael-smirnov: In my point of view, it is not (and this is mentioned in the definision of the term "have the same type and **domain**"). If data are homogeneous, then we can act with all the features similarly, we don't need to check the details about the types of each feature - it's a verbose simplification. If we allow some exceptions (like with domain), it may be unclear. I think that one needs to use homogeneous data if and only if it is claimed that all the features can be handled similarly.

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Reformulated with C++ terminology - copy-assignable and copy-constructible. Thanks!

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Changed `label` to `response`. Cannot see why it is incomplete. Even in the source mentioned by you, the regression task is described similar to my definition:
>We tend to refer to problems with a quantitative response as regression problems

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Reformulated

michael-smirnov:
Glossary updated and added examples for some terms. @rlnx @andreynv @outoftardis , please take a look.

source/elements/oneDAL/source/glossary.rst
andreynv: We potentially limit the scope of 'homogeneous data', only the data the belong to the same domain would be considered homogeneous. To me, the important aspect to keep in mind to differentiate between homo-/heterogeneous data - a set of ops we can do on the type: say, if we can apply multiplication, summation, comparison, etc to the data represented with the same data type, they are about the same, and domain is secondary and serves for the next level of checks on validity of the data or so during the compute stage or in error handling

source/elements/oneDAL/source/glossary.rst
```diff
@@ -13,92 +13,113 @@ Machine learning terms
...
-        On the :term:`training` stage, the model which finds the best dependency between input :term:`feature vector` and
-        :term:`labels<Label>` is selected. :term:`Label` is one of outputs can be made by :term:`model`
-        on the :term:`inference` stage.
+        A :term:`response` with :term:`categorical<Categorical feature>` or :term:`ordinal<Ordinal feature>`
+        values. This is an output in :term:`classification` and :term:`clustering` tasks.
+
+        Examples: the spam-detection task has a binary label indicating whether the email is spam or not.
 
     Model
         A result of :term:`training` machine-learning algorithm that stores information necessary to run
```
andreynv, line 75: by machine learning algorithm?

source/elements/oneDAL/source/glossary.rst
```diff
@@ -13,92 +13,113 @@ Machine learning terms
...
         :term:`inference` on a new :term:`dataset`.
 
+        Examples: in linear regression algorithm, the model contains weight values for each input feature
+        and single bias value.
+
     Observation
-        A vector of at least one :term:`feature` value and zero or more :term:`label<Label>` values.
```
andreynv: Just a feature vector and zero or more...

source/elements/oneDAL/source/glossary.rst
```diff
@@ -13,92 +13,113 @@ Machine learning terms
     Categorical feature
         A :term:`feature` with a discrete domain without ordering between values.
```
andreynv, line 14: to double-check: can categorical feature be ordinal, thus, ordinal and nominal are two classes of categorical features.  Also, double-check with Mikhail A. on another interval(?) class of features

source/elements/oneDAL/source/glossary.rst
```diff
@@ -13,92 +13,113 @@ Machine learning terms
...
-        :term:`continuous<Continuous feature>` :term:`labels<Label>` to
+        :term:`continuous<Continuous feature>` :term:`responses<Response>` to
         :term:`feature vectors<Feature vector>`.
 
+        Examples: prediction of the temperature based on weather conditions.
+
     Response
-        A value of a single :term:`label`. For example, response may be a flag
-        indicating that an observation belongs to a particular class.
+        A property of some real object or event which value is target to be predicted in the
```
andreynv, line 100: is target to be predicted - we need a help of Ekaterina to word it in a cleaner way

source/elements/oneDAL/source/glossary.rst
```diff
@@ -114,11 +135,15 @@ Machine learning terms
         A special property of the :term:`data format`: the data are stored as one contiguous memory block.
 
     Data format
-        The representation of internal structure of the data: this information is enough to efficiently access them.
```
andreynv: we can or can't introduce data layout but we need to understand differences between two defs

source/elements/oneDAL/source/glossary.rst
```diff
@@ -114,11 +135,15 @@ Machine learning terms
...
+        The representation of internal structure of the data.
+
+        Examples: data can be stored in array-of-structures or compressed-sparse-row format.
 
     Data type
         An integral or floating-point type which is used to physically store the data.
 
+        Examples: int32_t, float, double.
+
     Homogeneous data
```
andreynv, line 147: We also need to define heterogeneous data

source/elements/oneDAL/source/glossary.rst
```diff
@@ -130,9 +155,10 @@ Machine learning terms
         becomes available in parts over time.
 
     Reference-counted object
-        A |dal_short_name| object which stores the number of references to the unique implementation.
-        Such object has a lightweight copy constructor and assignment operator, which simply increases the number of references.
-        Implementation is automatically freed when the number of references becomes equal to zero.
+        A copy-constructible and copy-assignable |dal_short_name| object
```
andreynv, line 158: reference to the source that contains def of copy-constructible?

source/elements/oneDAL/source/glossary.rst
```diff
@@ -114,11 +135,15 @@ Machine learning terms
         A special property of the :term:`data format`: the data are stored as one contiguous memory block.
 
     Data format
-        The representation of internal structure of the data: this information is enough to efficiently access them.
+        The representation of internal structure of the data.
+
+        Examples: data can be stored in array-of-structures or compressed-sparse-row format.
 
     Data type
         An integral or floating-point type which is used to physically store the data.
```
andreynv, line 143: just represent or describe the data?

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Yes, I agree with you. It was my misunderstanding that domain defines the available operations too.

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Reformulated

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Reformulated

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Thanks, that was a mistake. Added terms for nominal, interval and ratio features

source/elements/oneDAL/source/glossary.rst
michael-smirnov: @outoftardis , please, help :)

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Added `data layout` term to show a difference between them

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Done

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Hmm... this is defined by the C++ standard, which is not freely available. I can add a reference to cppreference.com, which is a (non-official) reference for all terms and entities defined by C++ standard.
But do we really need that? This is a spec for C++/DPC++ library, so we can assume that the reader is familiar with C++ terms.

source/elements/oneDAL/source/glossary.rst
michael-smirnov: I think this term can be mentioned for both reasons. When we are talking about metadata, we mind "to describe and represent".  When we are talking about some variable which has a particular data type, we mean "to represent".

michael-smirnov:
@rlnx @andreynv @outoftardis glossary updated one more time. New terms added like `data layout`, `table`, `builder`, `accessor` etc. Some terms reworked.

source/elements/oneDAL/source/glossary.rst
outoftardis: we don't have any other style, yours is ok, but I suggest these changes: 
- remove the full stop
- make the word **Synonyms** bold
- use **Synonym** if there is only one

The same goes for examples/example.

source/elements/oneDAL/source/glossary.rst
outoftardis: sure, I'm reviewing the changes now and will add there a suggestion how to rephrase it

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Done

source/elements/oneDAL/source/glossary.rst
andreynv: okay, if the others think in the same direction

source/elements/oneDAL/source/glossary.rst
andreynv: type is used to store the data..but the data does not seem to be stored by type

andreynv:
overall it has the right content and looks good. The English is to be improved in some places. We use 'task' and 'problem' to describe machine learning. It makes sense to stick to one term, like problem or so. I may have few additional comments later this day

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,277 @@
...
+        An entiry which stores an information necessary to run :term:`inference`
+        on a new :term:`dataset`. Typically a result of :term:`training`
+        process.
+
+        **Example:** in linear regression algorithm, the model contains weight
+        values for each input feature and single bias value
+
+    Observation
+        A :term:`feature vector` and zero or more :term:`responses<Response>`.
+        Each of them can encode different objects or events.
```
outoftardis, line 101: each of them - each of what?

```suggestion
        A :term:`feature vector` and zero or more :term:`responses <Response>`.
        Each of them can encode different objects or events.
```

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,277 @@
...
+
+|dal_short_name| terms
+======================
+
+.. glossary::
+    :sorted:
+
+    Accessor
+        An |dal_short_name| concept of an object, which provides an access in the
+        special :term:`data format` to the data of another object.
```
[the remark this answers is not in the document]
outoftardis: also not sure about "concept of an object". what about "concept for an object"?

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,277 @@
...
+
+    Batch Mode
+        The computation mode for an algorithm in |dal_short_name|, where all the
+        data needed for computation is available at the start and fits the
+        memory of the device on which the computations are performed.
+
+    Builder
+        An |dal_short_name| concept of an object which encapsulates the creation
+        process of another object and provides an ability to create it
+        iteratively.
```
outoftardis, line 175: also same comment about concept as for accessor

```suggestion
        A |dal_short_name| concept of an object that encapsulates the creation
        process of another object and enables its iterative creation.
```

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,277 @@
...
+    Training set
+        A :term:`dataset` used at the :term:`training` stage to create a
+        :term:`model`.
+
+    Unsupervised learning
+        :term:`Training` process where :term:`training set` does not contain
+        information on dependencies between :term:`feature vectors<Feature
+        vector>` and :term:`responses<Response>`. The goal is to get a
+        :term:`model` matching :term:`feature vectors<Feature vector>` into
+        groups.
```
outoftardis, line 155: Not sure about dependencies. We don't have responses in unsupervised learning, right?
My suggestion:
```suggestion
        :term:`Training` process that uses a :term:`training set` with no
        :term:`responses <Response>`. The goal is to get a
        :term:`model` that allocates :term:`feature vectors <Feature vector>`  into
        groups.
```

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,277 @@
...
+        :term:`inference` stage.
+
+        **Synonym:** dependent variable
+
+    Supervised learning
+        :term:`Training` process with a :term:`dataset` containing information
+        on dependencies between :term:`features<Feature>` and
+        :term:`responses<Response>`. The goal is to get a :term:`model` of
+        dependencies between input :term:`feature vector` and
+        :term:`responses<Response>`.
```
outoftardis, line 139:
```suggestion
        :term:`Training` process that uses a :term:`dataset` with information
        about dependencies between :term:`features <Feature>` and
        :term:`responses <Response>`. The goal is to get a :term:`model` of
        dependencies between input :term:`feature vector` and
        :term:`responses <Response>`.
```

Also note that here the reference goes to `dataset`, while in unsupervised learning you refer `training set`.

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,277 @@
...
+    Ordinal feature
+        A :term:`categorical feature` with defined operations of equality an
+        ordering between values.
+
+        **Example:** student's grade
+
+    Ratio feature
+        A :term:`continuous feature` with defined operations of equality,
+        comparison, addition, substraction, multiplication and division.
+        Has a zero value element which means the absence of value.
```
outoftardis, line 114:
```suggestion
        A :term:`continuous feature` with defined operations of equality,
        comparison, addition, subtraction, multiplication, and division.
        Zero value element means the absence of any value.
```

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,277 @@
...
+
+    Observation
+        A :term:`feature vector` and zero or more :term:`responses<Response>`.
+        Each of them can encode different objects or events.
+
+        **Synonyms:** instance, sample
+
+    Ordinal feature
+        A :term:`categorical feature` with defined operations of equality an
+        ordering between values.
```
outoftardis, line 107:
```suggestion
        A :term:`categorical feature` with defined operations of equality and
        ordering between values.
```

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Good question about responses. At the training stage yes, we do not have them. But how to refer the outputs of the unsupervised task made at the inference stage? In my opinion, response can be used for supervised (regression+classification) tasks and unsupervised tasks. Label can be used for classification and clustering(unsupervised) tasks.
@rlnx , @andreynv , what is your opinion, can we use `response` for unsupervised tasks?

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,277 @@
...
+        memory of the device on which the computations are performed.
+
+    Builder
+        An |dal_short_name| concept of an object which encapsulates the creation
+        process of another object and provides an ability to create it
+        iteratively.
+
+    Contiguous data
+        A special property of the :term:`data format`: the data are stored as
+        one contiguous memory block.
```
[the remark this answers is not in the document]
michael-smirnov: Just because it is a single attribute of data format. It does not fully describe data format: data can be continuous, heterogeneous, stored in row-major layout with feature types (double, float, int). That's the full description of data format.

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,277 @@
...
+    Online Mode
+        The computation mode for an algorithm in |dal_short_name|, where the
+        data needed for computation becomes available in parts over time.
+
+    Reference-counted object
+        A copy-constructible and copy-assignable |dal_short_name| object which
+        stores the number of references to the unique implementation. Has
+        lightweight copy operations which increases the number of references.
+        An implementation is automatically freed when the number of references
+        becomes equal to zero.
```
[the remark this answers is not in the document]
michael-smirnov: > not clear to me what is meant here

Such objects share a unique implementation. This implementation stores the number of these objects. When you creating a new object referencing to implementation, this number increases. The object can be created by copy-constructor or assignment operator - "copy operations". Usually, they perform heavy and costly full-copy actions, but here the only thing we do is just incrementing one number.

source/elements/oneDAL/source/glossary.rst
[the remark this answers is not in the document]
outoftardis: would this be correct?

```
Data that is stored as one contiguous memory block.
One of the characteristics of a :term:`data format`.
```

source/elements/oneDAL/source/glossary.rst
andreynv: Unsupervised learning is about the case when we have feature vectors without responses and want to find some hidden patterns, e.g. how feature vectors are 'similar'. How we represent the outcome of this learning is secondary topic, e.g., we numerate the groups of the similar feature vectors or we have a set of rules that indicate connection between individual items/features in the data (if I buy this thing, I'll highly likely buy that thing)

michael-smirnov: Ok, thanks.

source/elements/oneDAL/source/glossary.rst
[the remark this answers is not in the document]
michael-smirnov: Yes, but I'd like to write "data that are stored..." because data is a plural noun. Correct me please, if I wrong.

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,277 @@
...
+
+        **Example:** A :term:`dataset` with 100
+        :term:`observations<Observation>` of 3 :term:`interval features<Interval
+        feature>`. The first two features have data type of float32, but the
+        third one of float64
+
+    Homogeneous data
+        A special property of the :term:`data format`: all the types of the
+        :term:`features<Feature>` are the same. This means that they have
+        similar :term:`data type` and defined set of available operations.
```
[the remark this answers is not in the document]
michael-smirnov: Reformulated.

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,277 @@
...
+
+    Data format
+        The representation of internal structure of the data.
+
+        **Examples:** data can be stored in array-of-structures or
+        compressed-sparse-row format
+
+    Data layout
+        A property :term:`data format` of :term:`contiguous data`. Describes an
+        order of elements inside contiguous data block.
```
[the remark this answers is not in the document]
michael-smirnov: Reformulated.

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Second statement removed. Detailed description is written for feature vector.

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,277 @@
...
+    Data type
+        An integral or floating-point type which is used to physically store the
+        data.
+
+        **Examples:** int32_t, float, double
+
+    Heterogeneous data
+        A property of the :term:`data format`, where :term:`features<Feature>`
+        have different :term:`data types<Data type>` and/or operations defined
+        between values.
```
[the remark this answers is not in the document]
michael-smirnov: Reformulated.

source/elements/oneDAL/source/glossary.rst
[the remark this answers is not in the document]
outoftardis: there's a controversy :) 
generally both `data are`/`data is` are ok as long as you use them consistently, but there's a tendency towards data as a singular mass noun (similar to information)
if you look at other places in this doc, you'll see data used as a singular

source/elements/oneDAL/source/glossary.rst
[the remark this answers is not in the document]
outoftardis:
```
Both copy operations defined for this object are lightweight, 
which means that each time a new object is created, only the number of references is increased by one.
```

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,275 @@
...
+        assigning :term:`labels <Label>` to :term:`feature vectors <Feature
+        vector>`.
+
+        **Examples:** predict what type of object is on the picture (a dog or a cat?),
+        predict whether or not an email is spam
+
+    Clustering
+        An :term:`unsupervised machine learning problem <Unsupervised learning>`
+        of grouping :term:`feature vectors <Feature vector>` into groups,
+        usually :term:`categorical <Categorical feature>`,
```
rlnx, line 36: What does "categorical groups" mean? The attempt to logically connect groups and categorical features seems irrelevant here

rlnx: "grouping into groups" sounds odd

rlnx: An unsupervised machine learning problem of grouping a set of feature vectors in a way that observations in the same group (called a cluster) are more similar to each other than to those in other clusters.

michael-smirnov: Rephrased a bit

source/elements/oneDAL/source/glossary.rst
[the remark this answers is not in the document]
michael-smirnov: Reformulated

source/elements/oneDAL/source/glossary.rst
michael-smirnov: Reformulated

source/elements/oneDAL/source/glossary.rst
```diff
@@ -0,0 +1,275 @@
...
+        :term:`responses<Response>` to :term:`feature vectors<Feature vector>`.
+
+        Examples: prediction of the temperature based on weather conditions.
+
+    Response
+        A property of some real object or event which value is target to be
+        predicted in the :term:`supervised learning` task. While the
+        :term:`feature` is an input in the machine learning task, the response
+        is one of the outputs can be made by the :term:`model` on the
+        :term:`inference` stage.
```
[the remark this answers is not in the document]
michael-smirnov: Reformulated

michael-smirnov:
@andreynv @rlnx @outoftardis I applied your comments, please look one more time. Compiled version is updated too, as usual.
@andreynv , I also extended terms of `table` and `accessor` with more details and added the term `metadata`. For builder, nothing comes to my mind about how I need to bring more details, because it is very common principle and potentially can be used to create not tables only.

source/elements/oneDAL/source/glossary.rst
```diff
@@ -219,6 +221,17 @@ Machine learning terms
         The object is immutable if it is not possible to change its state after
         creation.
 
+    Metadata
+        An information about logical and physical structure of an object. All
+        possible combinations of metadata values present the full set of
+        possible objects of a given type. Metadata do not expose an information
+        which is not a part of type definition, e.g. implementation details.
```
outoftardis, line 228:
```suggestion
    Metadata
        Information about logical and physical structure of an object. All
        possible combinations of metadata values present the full set of
        possible objects of a given type. Metadata do not expose information
        that is not a part of a type definition, e.g. implementation details.
```

PR #140, opened 2020-05-19
onemkl: update blas, memory model, fix warnings

mkrainiuk:


source/elements/oneMKL/source/domains/blas/asum.rst
```diff
@@ -58,15 +49,27 @@ asum
    where ``x`` is a vector with ``n`` elements.
 
 
+asum (BUFFER Version)
+---------------------
+
+.. container::
```
rscohn2, line 55: The .. container is an artifact of the translation from the original doc. It doesn't do anything. It's ok to commit this now, but we should do a cleanup pass (at the same time we have to unindent the contents). Same for :class: sectiontitle

source/elements/oneMKL/source/domains/blas/asum.rst
```diff
@@ -98,16 +100,84 @@ asum
...
 
    .. container:: parentlink
 
 
       **Parent topic:** :ref:`blas-level-1-routines`
-      
-
-
-.. |image0| image:: ../equations/GUID-684BB993-83CA-4605-BD49-E493806C1ee1.png
+.. |image0| image:: ../equations/GUID-4F76F5A1-251F-4AC0-A2E0-A3B4B6F39ee1.png
```
rscohn2, line 181: When I edit a doc with a GUID name image, I try to rename it to something more helpful.

source/elements/oneMKL/source/domains/blas/blas-level-1-routines.rst
```diff
@@ -0,0 +1,77 @@
...
+
+   .. container:: tablenoborder
+
+
+      .. list-table:: 
+         :header-rows: 1
+
+         * -     Routines
+           -     Description     
+         * -     \ `asum <asum.html>`__\
```
rscohn2, line 23: This also needs a cleanup pass. The link to html will not work in the PDF doc.  Asum should have  a label:
.. _asum:

And then you can refer to it with :ref:`asum`

I also prefer the simple table to .. list-table::

=======  ====
Routines      Description
=======  ====
:ref:`copy`     Copy vector  
:ref:`dot`       Dot product
=======  ====

source/elements/oneMKL/source/domains/blas/gemm_batch.rst
```diff
@@ -20,12 +24,12 @@ gemm_batch
       **Group API**
 
 
-      .. cpp:function::  void gemm_batch(queue &exec_queue,      buffer<transpose, 1> &transa_array, buffer<transpose,1>      &transb_array, buffer<std::int64_t,1> &m_array,      buffer<std::int64_t,1> &n_array, buffer<std::int64_t,1> &k_array,      buffer<T,1> alpha_array, buffer<T,1> &a_array,      buffer<std::int64_t,1> &lda_array, buffer<T,1> &b_array,      buffer<std::int64_t,1> ldb_array, buffer<T,1> &beta_array,      buffer<T,1> &c, buffer<std::int64_t,1> &ldc_array, std::int64_t      group_count, buffer<std::int64_t,1> &group_size_array)
```
rscohn2, line 23: cpp is the default, so .. function:: is enough

I have found that .. cpp:function:: does a poor job of formatting multi-line signatures and we should think about just using :: instead so we can have proper line-breaks

source/elements/oneMKL/source/domains/blas/gemm_ext.rst
```diff
@@ -69,7 +73,7 @@ gemm_ext
       .. container:: dlsyntaxpara
 
 
-         .. cpp:function::  void gemm_ext(queue &exec_queue, transpose         transa, transpose transb, offset offset_type, std::int64_t m,         std::int64_t n, std::int64_t k, Ts alpha, buffer<Ta,1> &a,         std::int64_t lda, Ta ao, buffer<Tb,1> &b, std::int64_t ldb, Tb         bo, Ts beta, buffer<Tc,1> &c, std::int64_t ldc, buffer<Tc,1>         &co)
+         .. cpp:function::  void onemkl::blas::gemm_ext(sycl::queue &queue, transpose         transa, transpose transb, offset offset_type, std::int64_t m,         std::int64_t n, std::int64_t k, Ts alpha, sycl::buffer<Ta,1> &a,         std::int64_t lda, Ta ao, sycl::buffer<Tb,1> &b, std::int64_t ldb, Tb         bo, Ts beta, sycl::buffer<Tc,1> &c, std::int64_t ldc, sycl::buffer<Tc,1>         &co)
```
rscohn2, line 76: I have not looked at the entire document, but you might want to have a section about namespaces. You could say that queue is always a sycl::queue so you won't have to put sycl:: everywhere. And it is probably better not to have onemkl:: namespace, but document that everything is in that namespace.

source/elements/oneMKL/source/domains/blas/asum.rst
mkrainiuk: Agree, I will add it to our todo list

source/elements/oneMKL/source/domains/blas/gemm_ext.rst
spencerpatty: our thought was to make the individual api's page definition copyable as best we can and as specific as we can without assumption of "using xxxxx" having been present.  We decided to add all of these namespaces in, in this one place.  Anywhere else it is referenced, it would not have all these details.  There is also a section in the Architecture chapter on namespaces.

source/elements/oneMKL/source/domains/blas/blas-level-1-routines.rst
spencerpatty: @rscohn2 FYI, our ultimate plan is to have the onemkl tags use the prefix  "_onemkl" to avoid collisions.  An extension would be for APIs in a given onemkl domain, to add "_<domain>" that is .. _onemkl_blas_asum  would be the appropriate tag here.  And   _onemkl_blas_copy  and _oneml_blas_dot ...

source/elements/oneMKL/source/domains/blas/asum.rst
spencerpatty: I think we have slowly come to understand the same thing (that .. container does not do anything)  We plan to clean this up at some point, but as you say, it will require lots of indentation changes...

source/elements/oneMKL/source/domains/blas/asum.rst
mkrainiuk: I will add it to our todo list

source/elements/oneMKL/source/domains/blas/blas-level-1-routines.rst
mkrainiuk: Good point, we definitely should use ref. Tables were generated with transition to rst, I think we will update them at some point as well.

source/elements/oneMKL/source/domains/blas/gemm_batch.rst
mkrainiuk: But will "::" block apply proper cpp syntax highlight?

rscohn2: It will do syntax highlighting, but not cross-referencing.  If you use .. parsed-literal::
You can do the cross-referencing manually

PR #151, opened 2020-05-27
Update oneDNN specification to 0.8

rsdubtso:
- Restructured the specification
- More detailed descriptions of execution and data models
- More general information on the API principles


source/elements/oneDNN/source/primitives/normalization.rst
```diff
@@ -0,0 +1,25 @@
+..
+  Copyright 2019-2020 Intel Corporation
+
+.. default-domain:: cpp
```
rscohn2, line 4: In source/conf/common.py we set the primary_domain to cpp so this isn't necessary. It's harmless to commit but you may want to delete them in the future.

rsdubtso: Understood. I was just adding them in case we get a heterogeneous API and the `primiary_domain` goes away.

rscohn2: OK. It is up to you.

Issue #147, opened 2020-05-25
[oneDAL] Glossary produces a lot of warnings


rlnx:
After merging #134, oneDAL part start producing the following warnings:

<details><summary>Build log</summary>

```
oneDAL\source\glossary.rst:18: WARNING: term feature not found in case sensitive match.made a reference to Feature instead.
oneDAL\source\glossary.rst:36: WARNING: term feature not found in case sensitive match.made a reference to Feature instead.
oneDAL\source\glossary.rst:43: WARNING: term model not found in case sensitive match.made a reference to Model
instead.
oneDAL\source\glossary.rst:43: WARNING: term dataset not found in case sensitive match.made a reference to Dataset instead.
oneDAL\source\glossary.rst:43: WARNING: term response not found in case sensitive match.made a reference to Response instead.
oneDAL\source\glossary.rst:50: WARNING: term dataset not found in case sensitive match.made a reference to Dataset instead.
oneDAL\source\glossary.rst:50: WARNING: term inference not found in case sensitive match.made a reference to Inference instead.
oneDAL\source\glossary.rst:54: WARNING: term continuous feature not found in case sensitive match.made a reference to Continuous feature instead.
oneDAL\source\glossary.rst:60: WARNING: term response not found in case sensitive match.made a reference to Response instead.
oneDAL\source\glossary.rst:60: WARNING: term classification not found in case sensitive match.made a reference
to Classification instead.
oneDAL\source\glossary.rst:60: WARNING: term clustering not found in case sensitive match.made a reference to Clustering instead.
oneDAL\source\glossary.rst:68: WARNING: term inference not found in case sensitive match.made a reference to Inference instead.
oneDAL\source\glossary.rst:68: WARNING: term dataset not found in case sensitive match.made a reference to Dataset instead.
oneDAL\source\glossary.rst:68: WARNING: term training not found in case sensitive match.made a reference to Training instead.
<rst_prolog>:20: WARNING: term categorical feature not found in case sensitive match.made a reference to Categorical feature instead.
oneDAL\source\glossary.rst:76: WARNING: term feature vector not found in case sensitive match.made a reference
to Feature vector instead.
oneDAL\source\glossary.rst:81: WARNING: term categorical feature not found in case sensitive match.made a reference to Categorical feature instead.
oneDAL\source\glossary.rst:91: WARNING: term continuous feature not found in case sensitive match.made a reference to Continuous feature instead.
[...]

PR #154, opened 2020-05-29
[oneDAL] build warnings fix

michael-smirnov:
This PR fixes problem #147 and all other warnings from oneDAL component:

- All warnings in glossary section fixed
- All warnings in data management section fixed
- Fixed bullet list indentations in introduction section

Warnings produced by case-sensitive checks are resolved this way:
1. When the term mentioned has the name exactly as in the glossary, it is started from the capital letter
2. If the term does not exactly matches to the related glossary section, it is started from lowercase letter.

Example: 
- To filter and normalize **Feature** values that are being extracted
- It is an in-memory or out-of-memory tabular view of data, where
table rows represent the **observations** and columns
represent the **features**.

Compiled version: https://michael-smirnov.github.io/oneapi-spec/elements/oneDAL/source/index.html

### DIFF
--- a/source/elements/oneDAL/source/data_management/index.rst
+++ b/source/elements/oneDAL/source/data_management/index.rst
@@ -56,7 +56,7 @@ Dataset
 --------
 
 The main data-related concept that |dal_short_name| works with is a
-:term:`dataset`. It is an in-memory or out-of-memory tabular view of data, where
+:term:`Dataset`. It is an in-memory or out-of-memory tabular view of data, where
 table rows represent the :term:`observations <Observation>` and columns
 represent the :term:`features <Feature>`.
 
@@ -82,7 +82,7 @@ example:
 Data source
 -----------
 
-Data source is a concept of an out-of-memory storage for a :term:`dataset`. It is
+Data source is a concept of an out-of-memory storage for a :term:`Dataset`. It is
 used at the data acquisition and data preparation stages for the following:
 
 - To extract datasets from external sources such as databases, files, remote
@@ -92,11 +92,11 @@ used at the data acquisition and data preparation stages for the following:
   the local memory, especially when processing with accelerators. A data source
   provides the ability to load data by batches and extracts it directly into the
   device's local memory. Therefore, a data source enables complex data analytics
-  scenarios, such as :term:`online computations <Online mode>`.
+  scenarios, such as :term:`online computations <Online Mode>`.
 
-- To filter and normalize :term:`feature` values that are being extracted.
+- To filter and normalize :term:`Feature` values that are being extracted.
 
-- To recover missing :term:`feature` values.
+- To recover missing :term:`Feature` values.
 
 - To detect :term:`outliers <Outlier>` and recover the abnormal data.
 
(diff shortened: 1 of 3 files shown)


michael-smirnov:
@rlnx @outoftardis what`s your opinion about this fix? Do we want to make it more consistent and refer to the terms always starting with capital or lowercase letter?

outoftardis:
I don't think it's a good idea to use capitalized terms in text. I provided a possible solution in the related issue.

michael-smirnov:
@outoftardis does it mean that I should not use capitalized references at all? There are some references to the topic sections, like "Dataset". When I somewhere else give a reference to this topic, it will be capitalized. Should I make lowercase aliases for such references?

rlnx:
> I don't think it's a good idea to use capitalized terms in text.

Agree with @outoftardis, capitalized word in the middle of the text seems odd to me.

> I provided a possible solution in the related issue.

Suppressing all the warnings related to the terms is also not very safe, as it could hide the real warning about missing term.

rlnx:
@michael-smirnov, the best solution, probably, is to create [custom role](https://protips.readthedocs.io/link-roles.html) for Sphinx that would automatically add reference to the capitalized term. It should allow using non-capitilized text, but insert terms as if it was capitilized.

```rst
:capterm:`feature`
```

rlnx:
Shouldn't be difficult to implement:

```py
from docutils import nodes

def setup(app):
    app.add_role('capterm', capterm_role)

def capterm_role(name, rawtext, text, lineno, inliner, options={}, content=[]):
    node = nodes.term(rawtext, text.title()) 
    return [node], []
```

This can be added [here](https://github.com/oneapi-src/oneAPI-spec/blob/fb5cecb664bfdf7cd4b48351fc1c929251d6e062/source/elements/oneDAL/dalapi/extension.py#L278). I'm not 100% sure the code snippet above is working as expected 😄. I used `docutils` reference from [here](http://code.nabla.net/doc/docutils/api/docutils/nodes/docutils.nodes.term.html#docutils.nodes.term).

michael-smirnov:
@rlnx @outoftardis you don't answer the main question - how to deal with links to a document sections? They are capitalized, for example:

> Logically, a table is a **Dataset** with n rows and p columns. Each row represents an observation and each column is a **feature** of a dataset.

Here **Dataset** is a link to the corresponding section, but **feature** is a link to the glossary term. 
**Do you think we should also fix links to the sections and do not use capitalized links at all?**

michael-smirnov:
Here we also have a term **dataset**. Do we want to distinguish cases when I refer to the dataset section and to the glossary term "dataset"? Or we should introduce some simple term, e.g. reference always to the section if exists.

rlnx:
> @rlnx @outoftardis you don't answer the main question - how to deal with links to a document sections? They are capitalized, for example:

@michael-smirnov, I prefer to avoid the capitalized links at all [as I said previously](https://github.com/oneapi-src/oneAPI-spec/pull/154#issuecomment-636055374).


> Here we also have a term dataset. Do we want to distinguish cases when I refer to the dataset section and to the glossary term "dataset"? Or we should introduce some simple rule, e.g. reference always to the section if exists.

Simple rule would be to refer the **Dataset** section always.

rlnx:
Implemented custom role approach in #160

michael-smirnov:
Closed because the better approach is proposed in #160

PR #158, opened 2020-06-01
oneDNN: work around Sphinx type lookup bug

rsdubtso:
https://github.com/sphinx-doc/sphinx/issues/7740

### DIFF
--- a/source/elements/oneDNN/include/dnnl.hpp
+++ b/source/elements/oneDNN/include/dnnl.hpp
@@ -126,10 +126,10 @@ struct primitive {
     /// #dnnl::primitive_desc_base::query_md(#query::exec_arg_md, index)
     /// unless using dynamic shapes (see #DNNL_RUNTIME_DIM_VAL).
     ///
-    /// @param stream Stream object. The stream must belong to the same engine
+    /// @param astream Stream object. The stream must belong to the same engine
     ///     as the primitive.
     /// @param args Arguments map.
-    void execute(const stream &stream,
+    void execute(const stream &astream,
             const std::unordered_map<int, memory> &args) const;
 
     /// Executes computations specified by the primitive in a specified stream.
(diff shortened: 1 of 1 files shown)


source/elements/oneDNN/include/dnnl.hpp
```diff
@@ -1696,24 +1694,24 @@ struct reorder : public primitive {
...
     ///
-    /// @param stream Stream object. The stream must belong to the same engine
+    /// @param astream Stream object. The stream must belong to the same engine
     ///               as the primitive.
     /// @param src Source memory object.
     /// @param dst Destination memory object.
     /// @param deps Vector of SYCL events that the execution should depend on.
     ///
     /// @returns SYCL event that corresponds to the SYCL queue underlying the
     ///          @p stream.
```
emfomenk, line 1712:
```suggestion
    ///          @p astream.
```

source/elements/oneDNN/include/dnnl.hpp
```diff
@@ -541,17 +541,15 @@ struct engine {
 /// @{
 
 /// A container for stream attributes.
-struct stream_attr : public handle<dnnl_stream_attr_t> {
-    using handle::handle;
-
+struct stream_attr {
```
emfomenk, line 544: why?

rsdubtso: This header file is not supposed to have any mention of `handle<T>` and, especially, should not have any references to the C API, as this is implementation-specific.

source/elements/oneDNN/include/dnnl.hpp
rsdubtso: Fixed. Thanks!

source/elements/oneDNN/include/dnnl.hpp
emfomenk: oh, my bad, i didn't notice we don't have `handle<T>` in other places. sorry for the noise.

PR #161, opened 2020-06-05
[DAL] Lowercase references in the text

michael-smirnov:
This PR proposes a new custom rule `txtref`. As a standard `ref` rule, this provides references to some locations in a document.

Consider next reference declaration:
```rst
:: _some_ref:
Some reference
```
- `:ref:'some_ref'` will result to the text after `some_ref` declaration: `Some reference`
- `txtref:'some_ref'` will result to `some ref` and points to the same destination as a `ref` clause

**Rules of `txtref` role** 
1. **:txtref:'some_text'** will create a link to `some_text` destination and results to the original text of the link with delimiter characters (`-` and `_`) replaced by spaces:
   - :txtref:'some_text' -> some text
   -  :txtref:'some-text' -> some text
   - :txtref:'some_another-text' -> some another text
2. **:txtref:'Table_API'** will create a link to `Table_API` and results to `table API`, replacing the first capital letter by lowercase one
3. **:txtref:'some text link <some_text>'** will act the same as `:ref:` rule poining to some_text link and aliasing it by "some text link" text.
4. **:txtref:<some_object>s** -> someobjects, with reference to some_object destination. It will help to shorten references like "Iterative construction of a :ref:`table` from another :ref:`tables <table>`"

### DIFF
--- a/source/elements/oneDAL/dalapi/extension.py
+++ b/source/elements/oneDAL/dalapi/extension.py
@@ -280,6 +280,7 @@ def setup(app):
     ctx = Context(app)
 
     app.add_role('capterm', roles.capterm_role)
+    app.add_role('txtref', roles.txtref_role)
 
     app.add_directive('onedal_class', directives.ClassDirective(ctx))
     app.add_directive('onedal_func', directives.FunctionDirective(ctx))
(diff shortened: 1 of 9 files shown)


michael-smirnov:
@outoftardis @rlnx Here I also fixed the links in data management section that started from capital letter. And changed all occurrences of :ref: role to the :txtref:.
I'm not proposing to avoid the use of :ref: role in favor of :txtref:, or use the latter always.
But I think this role it is more 'predictable' since the link looks almost the same as you wrote, while :ref: role changes it by the text after link declaration.

PR #162, opened 2020-06-19
versioning proposal

rscohn2:
I am getting questions on how we manage versioning after 1.0. Here is recommendation/rationale. Using a PR for feedback.

### DIFF
--- a/doc/versioning.rst
+++ b/doc/versioning.rst
@@ -59,6 +59,34 @@ version is effectively an increase in minor version.
 Increases in major version of an element are coordinated with the
 overall specification and are expected to be rare.
 
+Example
+-------
+
+Initial release:
+
+| oneAPI 1.0
+| oneMKL 1.0
+| oneDAL 1.0
+
+oneMKL adds some API's, oneDAL does not change:
+
+| oneAPI 1.1
+| oneMKL 1.1
+| oneDAL 1.0
+
+oneDAL adds API's, oneMKL does not change:
+
+| oneAPI 1.2
+| oneMKL 1.1
+| oneDAL 1.1
+
+oneMKL removes API's:
+
+| oneAPI 2.0
+| oneMKL 2.0
+| oneDAL 1.1
+
+
 Relationship of Implementation and Specification Versions
 =========================================================
 
--- a/doc/versioning.rst
+++ b/doc/versioning.rst
@@ -16,25 +16,26 @@ Purpose
 =======
 
 oneAPI specification is primarily a description of APIs. We use
-`Semantic Versioning`_ so oneAPI users and implementors can understand
-the compatibility of specification releases and implementations of
-specifications. It covers cases such as adding interfaces, removing
-interfaces, and making non-compatible changes to existing
-interfaces. The versioning does not in general provide information on
-changes in the document that are not related to interfaces, such as
-document editing changes.
+`Semantic Versioning`_ to communicate functionality additions and
+compatibility to users and implementors.  It covers cases such as
+adding interfaces, removing interfaces, and making non-compatible
+changes to existing interfaces. The versioning does not in general
+provide information on changes in the document that are not related to
+interfaces, such as document editing changes.
 
 
 Semantic Versioning
 ===================
 
 A summary is included below and based on `Semantic Versioning`_. Rule
-3 is augmented with the definition from `Semantic Versioning for
+1 is augmented to include major functionality additions. Rule 3 is
+augmented with the definition from `Semantic Versioning for
 Documents`_.
 
   Given a version number MAJOR.MINOR.PATCH, increment the:
 
-  1. MAJOR version when you make incompatible API changes,
+  1. MAJOR version when you make incompatible API changes or add major
+     new functionality,
      
   2. MINOR version when you add functionality in a backwards
      compatible manner, and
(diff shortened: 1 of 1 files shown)


doc/versioning.rst
```diff
@@ -0,0 +1,69 @@
...
+  It is customary and preferred. Year would be valuable if product
+  2025 always implemented spec 2025. Since we cannot guarantee the
+  relations, no benefit to use year.
+
+Whole spec
+===========
+
+Versioning scheme?
+
+  semantic versioning. If any component bumps major version, then
```
spencerpatty, line 37: Are we talking about the oneAPI Spec versioning here?  and are the components = oneaPI elements in spec?  The way I read this statement is that the onAPI spec moves forward with any major or minor update from any element.  

Start here:
  oneAPI 1.0
  oneMKL 1.0
  oneDAL 1.0

oneMKL updates triggering an update to oneAPI
oneAPI 1.1
oneMKL 1.1
oneDAL 1.0

What about if two elements increment at roughly the same time?  Does oneAPI move forward twice?
What does it look like if an element increments major, then in same time, another element increments a minor tick?

I guess these issues may be what you are referring to in the two paragraphs lower:

> Whole spec semantic simplies talking about and assembling
>   collections of software, at the expense of freedom of the invidiual
>   components. To be useful, it requires some discipline when
>   individual components can bump major versions, so it happens on a
>   manageable cadence.

rsdubtso: I read this as if one component bumps minor/major version, then it triggers a bump in all the components as well as the whole spec. This reading (whether accurate or not) makes sense to me...

rscohn2: Complete rewrite to address Spencer's comments. I also talked to @andreynv --it is conceptually compatible with the library wg decision, but some of the details are different. We will work it out offline.

rsdubtso: @rscohn2 : can you please clarify why the whole spec needs a version number?

rscohn2: Is your question: 
* Why do we need semantic versioning?
* Why do we need the concept of a single version instead of just talking about individual element versions?

I believe most people will be only concerned about the version of an individual element so a oneAPI version will not be relevant. However, I think there is a benefit in being able to say that a program is compatible with oneAPI spec *X* instead of saying the individual element versions, or a release of a oneAPI product.

rsdubtso: My question is "*Why do we need the concept of a single version instead of just talking about individual element versions?*"

> I believe most people will be only concerned about the version of an individual element so a oneAPI version will not be relevant. However, I think there is a benefit in being able to say that a program is compatible with oneAPI spec X instead of saying the individual element versions, or a release of a oneAPI product.

I'm not sure the benefit because a component may continue being backward compatible across spec versions 1.x and 2.x, so this is a source of confusion. The whole spec version number does not say anything on it's own because it is necessary to know which component versions a oneAPI spec version corresponds to. 

I am not suggesting to keep version numbers in sync. This approach has its own problems and I'm not suggesting to consider it.

I am suggesting to not have the whole spec version, though...

However, I'm suggesting we don't require

rscohn2:
I got some feedback that breaking compatibility would be very rare so we are likely to be at 1.x for a very long time.  Making the major version increase on  major new functionality will make the major version number more useful. OpenMP1, ..., OpenMP5 was cited as an example, where they have not broken compatibility and use the number to indicate new functionality. 

Defining major functionality will be a judgement call.  As calibration, I think adding USM support would be major functionality, as it has impact on most aspects of oneAPI.

@rsdubtso is objecting to versioning the spec altogether. I believe the argument is that compatibility of overall spec is not that useful.

I just pushed a commit that redefines major to indicate major new functionality. I kept incompatible. Most likely, major functionality would be the driver for bumping major number, and incompatible changes might happen at the same time.  @rsdubtso, how do you feel about versioning now?

doc/versioning.rst
emfomenk: I second @rsdubtso opinion on that oneAPI spec version doesn't bring much value while introducing yet another number related to the library, which is confusing and quite annoying. I very like the suggestion to not have a version for the whole spec at all.

Say with oneDNN we will have (**4** versions for essentially **2** concepts -- spec and implementation):
- oneDNN version of the package (2021.zzz)
- oneDNN library version (2.0.0 or whatever)
- oneDNN spec version (say 1.2.0)
- Related oneAPI spec version (say 2.1.1)

Most applications that use oneAPI will most likely depend on one or two components, and not the whole oneAPI stack. So, they probably don't care much if oneAPI spec is 2 or 3 or whatnot. Moreover, if they do, they need to state something like: 
> _"Our software relies on oneAPI spec version X, more specifically on oneDNN spec version Y and oneTBB spec version Z. So our software might be compatible with oneAPI spec version X+1 if this version doesn't increase the spec versions of oneDNN and oneTBB"_.

This at least sounds weird...

> I got some feedback that breaking compatibility would be very rare so we are likely to be at 1.x for a very long time.

Since oneAPI consists of many-many components (9?, more will join?), that roughly means that the frequency of need to change the major version would be (rare / # of components), which could be not that rare.

rsdubtso:
@rscohn2 if major version can be incremented by a feature addition, then it is not semantic in my reading of the definition. Semantic versioning suggests that a new feature requires the minor version to be incremented if the change did not break backwards compatibility.

doc/versioning.rst
rscohn2: I don't think we are going to converge so I will schedule a call with spec leads + @emfomenk.

PR #166, opened 2020-06-29
Update conf.py

aregm:
Add an option to leave the chapter grouping on by default.


rscohn2:
Navigation works, better. Thanks!

PR #173, opened 2020-07-27
oneVPL spec v0.9 update

IgorBelyakovIntel:


source/elements/oneVPL/source/index.rst
```diff
@@ -3,313 +3,38 @@
...
-Structs
--------
-
-.. include:: VPL_structs.inc.rst
-
-Functions
----------------
-
-.. include:: VPL_functions.inc.rst
+.. toctree::
```
rscohn2, line 32: Here is the full spec rendered as HTML: http://ansatnuc02.an.intel.com/rscohn1-spec/oneapi-spec/build/html/elements/oneVPL/source/index.html

Look at VPL table of contents in the left pane. You can only the versions section.  You cannot navigate to archtecture, or any of the other sections. Looking at your index.html, you have set the TOC as hidden, and have nested it inside the versioning section.

I suggest you move the versioning section into a separate file, add it to the TOC, and then make the TOC unhidden.

rscohn2: Kristal will do the change in a separate commit. I will merge as-is.

PR #197, opened 2020-08-22
[DAL] Changed include paths, namespace name updated

michael-smirnov:
- include directory path changed from `include\onedal` to `include\oneapi\dal`
- namespace `onedal` changed to `oneapi::dal`


andreynv (approved):
overall, looks good. The only q-do we make sure we did not miss other places which contained 'onedal'?

michael-smirnov:
@andreynv, yes, I checked that. Now `onedal` present only in sphinx rule names

PR #232, opened 2020-09-10
[oneMKL][DFT][LAPACK] add some last changes for oneAPI 1.0 spec

spencerpatty:
DFT -- update compute_forward and compute_backward apis now that we have a more general knowledge of what they could look like

LAPACK -- update LAPACK specific exceptions in each api.


spencerpatty:
Note that the "Adjust Version" commit and the merge from upstream/master are pulling this fork into alignment with the oneAPI-spec master branch.  They should have no impact at all for this pull request.  The last two commits are the ones that are of interest.  As part of review with we need to make sure that the adjust version commit does not do something funny :)

spencerpatty:
> Note that the "Adjust Version" commit and the merge from upstream/master are pulling this fork into alignment with the oneAPI-spec master branch. They should have no impact at all for this pull request. The last two commits are the ones that are of interest. As part of review with we need to make sure that the adjust version commit does not do something funny :)

i have validated that it looks as we expect.  No issues here.  The two commits of interest are also looking good.  Overall I see it as ready to merge, but would like feedback from someone else. :)

rscohn2 (approved):
Re-using a branch that has already been merged leads to problems for me, too. I usually delete the old branch and create a new one. If the extra commits bother you, then it is possible to cherry pick the commits you want into a new branch. If you don't care then feel free to merge this PR.

PR #247, opened 2020-09-24
add reference to info_namespace for tbb

rscohn2:
@ivankochin: Please check this
standalone tbb spec (aux-interfaces.rst) and integrated (nested-aux-interfaces.rst) have separate top-level TOC. Add a missing file.


ivankochin (approved):
Oh, I didn't know that.
Thanks for fixing this issue!

PR #276, opened 2020-12-15
Disallow antidependencies between oneMKL APIs

petercad:
Small clarification to disallow calling a oneMKL computational routine with a dependency on a future oneMKL API call (can be done in a few ways). This allows implementations to claim state and resources in an asynchronous routine at the time of the first call without worrying about deadlocks.


mkrainiuk (approved):

petercad:
Thanks, @mkrainiuk! I don't have write access -- could you merge the PR for me?

PR #284, opened 2021-02-10
Changes for oneART publishing

rscohn2:
@dmnemshi : Please look over text, but not formatting changes


dmnemshi:
The text looks good.

PR #300, opened 2021-03-10
[oneDPL] Add section about how to pass data to algorithms

capatober:
It clarifies how user can pass data to oneDPL algorithms

### DIFF
--- a/source/elements/oneDPL/source/pstl.rst
+++ b/source/elements/oneDPL/source/pstl.rst
@@ -139,6 +139,13 @@ as the template argument, otherwise unspecified.
 Return a policy object constructed from ``policy``, with a new kernel name provided as the template
 argument. If no policy object is provided, the new policy is constructed from ``dpcpp_default``.
 
+Pass Data to Algorithms
+++++++++++++++++++++++++
+You can use one of the following ways to pass data to an algorithm executed with a DPC++ policy:
+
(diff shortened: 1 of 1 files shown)


akukanov:
I do not like it. The wording would be good for a user manual, but not good for the specification which should be more formal and oriented towards implementers, not users. The placement of this in the middle of API description is also not quite right.

capatober:
@akukanov, thanks for the feedback!
Do you mean the wording and placement only or the thing that we need this changes in the Spec as well?
If it's just about wording then would the following wording be better?
```
There are 2 forms of data that can be passed to oneDPL algorithms executed with a DPC++ policy:
...
Passing others as a data to algorithms executed with a DPC++ policy leads to undefined behavior.
```

I've placed it here because the Buffer wrappers section is mentioned in the changes and it seemed as a good idea to place the changes before the Buffer wrappers section. It looks like:
```
1. Policies.
2. Forms of data that are accepted by algorithms.
2.a. or 3. Buffer wrappers
```
I also see another way:
```
1. Algorithms and data accepted by algorithms.
2. Specific API
2.a. Policies
2.b. Buffer wrappers
```
Is it better?

rscohn2:
I am making a release. Can you merge or close this?

rscohn2:
I am closing because this has been inactive. You can reopen if you are still working on it.

Issue #304, opened 2021-03-23
namespace oneapi::std is a bad idea


akukanov:
oneDPL specifications v1.0 says (https://spec.oneapi.com/versions/1.0/elements/oneDPL/source/common.html) that "oneDPL uses ``namespace oneapi::std`` for the subset of the standard C++ library for kernels". In practice, this turned out to be a bad idea, because this namespace hides the standard ``namespace std`` for any code within ``namespace oneapi``, i.e. for all oneAPI libraries, resulting in the need to use double colons for the standard classes & functons, Because of this issue, Intel's implementation of oneDPL does not yet provide this namespace.

The recommendation is to remove ``namespace oneapi::std`` from the specification. It's an incompatible change technically, but since there is no implementation providing this namespace, the removal should have no impact on any real code.

rscohn2:
Out of an abundance of caution, I asked Alexey to post something here so other teams could comment. I agree with Alexey that we should proceed.

andreynv:
The recommendation makes sense to me

akukanov:
Fixed by PR https://github.com/oneapi-src/oneAPI-spec/pull/305

PR #305, opened 2021-03-23
Remove namespace oneapi::std from oneDPL

akukanov:
A specification patch that fixes https://github.com/oneapi-src/oneAPI-spec/issues/304

### DIFF
--- a/source/elements/oneDPL/source/common.rst
+++ b/source/elements/oneDPL/source/common.rst
@@ -5,7 +5,7 @@
 Namespaces
 ----------
 
-oneDPL uses :code:`namespace oneapi::std` for the subset of the standard C++ library for kernels,
-and uses :code:`namespace oneapi::dpl` for other functionality including Parallel STL algorithms,
-oneDPL execution policies, and extensions.
+oneDPL uses :code:`namespace oneapi::dpl` for all its functionality including Parallel STL algorithms,
+oneDPL execution policies, etc. For the subset of the standard C++ library for kernels, the standard class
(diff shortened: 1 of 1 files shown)


source/elements/oneDPL/source/common.rst
```diff
@@ -5,7 +5,7 @@
 Namespaces
 ----------
 
-oneDPL uses :code:`namespace oneapi::std` for the subset of the standard C++ library for kernels,
-and uses :code:`namespace oneapi::dpl` for other functionality including Parallel STL algorithms,
-oneDPL execution policies, and extensions.
+oneDPL uses :code:`namespace oneapi::dpl` for all its functionality including Parallel STL algorithms,
+oneDPL execution policies, etc. For the subset of the standard C++ library for kernels, the standard class
+and function names aer also aliased in :code:`namespace oneapi::dpl`.
```
rarutyun, line 10: probably it would be better to say "oneDPL uses :code:`namespace oneapi::dpl` for all its functionality including Parallel STL algorithms, oneDPL execution policies and the subset of the standard C++ library for kernels: the standard class and function names

rarutyun: aer -> are if the previous suggestion wouldn't be applied

akukanov: Thanks for finding the typo. I think having a separate sentence about the kernel subset is easier read (due to shorter sentences) and puts focus on the specifics of the subset.

PR #310, opened 2021-04-06
make watermark optional

rscohn2:
draft watermark is enabled by setting environment variable

    ONEAPI_DRAFT=true

for sphinx

### DIFF
--- a/ubuntu-packages.txt
+++ b/ubuntu-packages.txt
@@ -4,6 +4,7 @@ git
 graphviz
 inkscape
 latexmk
+plantuml
 texlive-fonts-extra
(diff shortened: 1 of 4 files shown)


ubuntu-packages.txt
```diff
@@ -4,7 +4,6 @@ git
 graphviz
 inkscape
 latexmk
-plantuml
```
IgorBelyakovIntel, line 7: I think we need to have `plantuml` here. It is required for the python binding.

rscohn2: I put plantuml back to how it was originally

PR #317, opened 2021-04-09
fix release notes

rscohn2:


rscohn2:
Changes were put into other commits

PR #330, opened 2021-05-12
Update oneVPL to 2.4.0

akwrobel:


rscohn2:
My intention was to add release notes for the latest and leave older release notes. See here for an example
https://spec.oneapi.com/releases/index.html#id2

akwrobel:
@rscohn2 This was one of my opens to @IgorBelyakovIntel with this update and the request was replacement.  I don't have any objections to this new formatting, and I'll push a change inline with this request shortly

rscohn2:
If someone has a strong preference for replacement over cumulative, then let me know so we can discuss it.

akwrobel:
@rscohn2 No strong opposition.  Igor indicated I should check with you if there were any questions.  
https://github.com/oneapi-src/oneAPI-spec/pull/331

PR #335, opened 2021-06-15
Amend and force-push to publish to GitHub Pages

### LINKED ISSUE #334: GitHub Pages keeps increasing repository size
This is basically a duplicate of https://github.com/camptocamp/ngeo/issues/2815

Each time GitHub Actions publishes to the gh-pages branch, a new commit is created containing build files, including binary files such as the large (15 MB) PDF specification. This makes the repository grow in size quite rapidly, and threatens to reach quotas and to make the repository slower/unworkable in the mid to longer term.

The current size of the repository is reported to be 181,246 KB by the GitHub API.

I think we need to rewrite the history of the gh-pages branch to clean the repository (once `git-gc` decides to remove all the large unreachable objects on the server), and to rethink how we manage publishing to GH Pages (amend and force push to a single commit on gh-pages every time? disable GH Pages?)

hdelassus:
This commit avoids making the repository grow in size too fast by
overwriting the gh-pages branch on every build. This commit is not
enough in itself. To decrease the repository size, we also need to do
something like cherry-pick the latest gh-pages into a new orphan branch,
then force-push this branch to gh-pages. Newly unreachable objects
should be slowly garbage-collected over time.

(Partially) Fixes #334.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,5 +44,5 @@ jobs:
           git config user.name github-actions
           git config user.email github-actions@github.com
           git add .
-          git commit -m "Update from github actions"
-          git push
+          git commit --amend -m "Update from github actions"
+          git push --force origin gh-pages


rscohn2:
Thanks.

I squashed all the changes after branching for gh-pages to clean up old history.

PR #336, opened 2021-06-15
[oneDAL] Update dalapi tool

rlnx:
Dalapi tool source code is synchronized with the master of oneDAL repo.

### DIFF
--- a/source/elements/oneDAL/dalapi/generator.py
+++ b/source/elements/oneDAL/dalapi/generator.py
@@ -39,7 +39,6 @@ class RstBuilder(object):
         assert declaration
         assert parent_fully_qualified_name
         fake_parent_namespace = '_'.join(parent_fully_qualified_name.split('::'))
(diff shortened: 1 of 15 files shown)


source/elements/oneDAL/dalapi/generator.py
```diff
@@ -39,7 +39,6 @@ def add_property_member(self, declaration: str, parent_fully_qualified_name: str
         assert declaration
         assert parent_fully_qualified_name
         fake_parent_namespace = '_'.join(parent_fully_qualified_name.split('::'))
```
michael-smirnov, line 41: nip: unused variable

michael-smirnov (approved):

source/elements/oneDAL/dalapi/generator.py
rlnx: Should be fixed in the oneDAL master and than synchronized again

PR #359, opened 2021-08-30
[oneTBB] task_group contructor acepting task_group_context

anton-potapov:
- added `task_group` constructor accepting `task_group_context`


anton-potapov:
@alexey-katranov , @ep

PR #361, opened 2021-08-30
[oneVPL] Update to oneVPL Spec v2.5

akwrobel:


releases/index.rst
```diff
@@ -185,6 +188,33 @@ oneVPL
...
+* External Buffer ID of mfxExtVideoSignalInfo and mfxExtMasteringDisplayColourVolume for video processing.
+* New MFX_WRN_ALLOC_TIMEOUT_EXPIRED return status. Indicates that all surfaces are currently in use and timeout set by mfxExtAllocationHints for allocation of new surfaces through functions GetSurfaceForXXX expired.
+* Introduced universal temporal layering structure.
+* Added MFX_RESOURCE_VA_SURFACE_PTR and MFX_RESOURCE_VA_BUFFER_PTR enumerators.
+* Introduced segmentation interface for AV1e, including ext-buffers and enums.
+* Introduced planar I422 and I210 FourCC codes.
+
+Bug Fixes:
+
+* Dispatcher: Removed /etc/ld.so.cache from oneVPL search order.
```
rscohn2, line 211: Is search order part of the spec, or only visible in the implementation? If it is not part of the spec, then it should not be  mentioned here.

akwrobel: Hi Robert, 
The load ordering for the dispatcher is very important, and is critical in understanding what mode the dispatcher is loading in, legacy vs current.  This is explicitly called out in the spec on the oneVPL session section.

PR #363, opened 2021-09-02
[oneMKL][spblas] updated sparse::gemm routines spec, which adds supporting matrix storage layout

baeseung-intel:
**Goal:**
for `sparse::gemm()` functionality, which is sparse matrix-dense matrix product, support dense matrix storage layout of `column-major` as well as `row-major`.

**Details:**
Currently, there is no parameter related to dense matrix storage layout in `spares::gemm()` API and assumes the dense matrices are in `row-major` matrix storage layout.  In this PR, I propose to support below features:
- support `layout` of dense matrices of matrix `B` and `C` for both `row-major` and `column-major`.
- add matrix operation parameter for `B` matrix, so that we can handle `transposed` case of matrix `B` as well as `non-transpose` case.
- introduce a new enum class for `layout` information as `oneapi::mkl::layout`, which consists of `row_major` and `col_major`. 
 
**Motivation:**
As I mentioned above, current `sparse::gemm()` API-spec doesn't have parameters for dense matrix storage layout information and and matrix operation on B, which supports only `row-major` layout and `non-transpose` operation on matrix `B`. So if a user needs to call `sparse::gemm()` with `column-major` layout dense matrices and/or `transpose` operation on matrix `B`, the user has to do extra works before and/or after using `sparse::gemm()` for their usage. 

By updating this proposed `sparse::gemm()` API, we can support directly for such cases without requiring additional steps. Also, as part of this proposal, I propose another enum class for `layout` information as well.

### DIFF
--- a/source/elements/oneMKL/source/architecture/api_design.inc.rst
+++ b/source/elements/oneMKL/source/architecture/api_design.inc.rst
@@ -229,6 +229,32 @@ Each enumeration value comes with two names: A single-character name (the tradit
               -  Index arrays for an input matrix are provided using one-based (Fortran style) index values.  That is, indices start at 1.
 
 
+      .. _onemkl_enum_layout:
+
+      .. rubric:: layout
+         :name: layout
+         :class: sectiontitle
+
+      The ``layout`` type specifies how a dense matrix ``A`` with leading dimension ``lda`` is stored as one dimensional array in memory. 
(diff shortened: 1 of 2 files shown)


rscohn2:
@spencerpatty : Do you want this in 1.1?

spencerpatty:
yes I think it is probably a good idea to put it in 1.1.

spencerpatty (approved):
I approve of this change.

PR #372, opened 2021-09-22
[oneTBB] Function object iso cpp requirements 

anton-potapov:
Fixed wording on user defined function to match ISO c++ standard named requirement `[function.objects]`
@alexandraepan could you please take a look ?


alexandraepan (approved):
good job!

PR #393, opened 2021-12-03
[oneVPL] Update to oneVPL Spec v2.6

akwrobel:


releases/index.rst
```diff
@@ -319,6 +319,41 @@ oneVPL
 Release Notes
 -------------
 
+oneVPL v2.6.0
```
rscohn2, line 322: Can you create an entry in the table for 2.6.0 at line 309 above?

rscohn2 (changes requested):

rscohn2 (approved):

PR #394, opened 2021-12-08
[oneTBB] Add the Core Types preference API to the spec

ivankochin:
It was decided the Core Types preference API can be added to the oneTBB spec reference. This interface allows setting the preferred core type for the threads within a certain arena on the system with hybrid processors.

### DIFF
--- a/source/elements/oneTBB/source/info_namespace.rst
+++ b/source/elements/oneTBB/source/info_namespace.rst
@@ -16,8 +16,13 @@ Interfaces to query information about execution environment.
     namespace oneapi {
     namespace tbb {
         using numa_node_id = /*implementation-defined*/;
+        using core_type_id = /*implementation-defined*/;
+
         namespace info {
             std::vector<numa_node_id> numa_nodes();
+            std::vector<core_type_id> core_types();
+
+            int default_concurrency(task_arena::constraints c);
             int default_concurrency(numa_node_id id = oneapi::tbb::task_arena::automatic);
         }
     } // namespace tbb
@@ -38,6 +43,19 @@ Functions
     .. note::
         If error occurs during system topology parsing, returns vector containing single element
         that equals to ``task_arena::automatic``.
+
+.. cpp:function:: std::vector<core_type_id> core_types()
+
+    Returns the vector of integral indexes that indicate available core types.
+    The indexes are sorted from the least performant to the most performant core type.
+
+    .. note::
+        If error occurs during system topology parsing, returns vector containing single element
+        that equals to ``task_arena::automatic``.
+
+.. cpp:function:: int default_concurrency(task_arena::constraints c)
+
+    Returns concurrency level for the given constraints.
 
 .. cpp:function:: int default_concurrency(numa_node_id id = oneapi::tbb::task_arena::automatic)
 
--- a/source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
+++ b/source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
@@ -159,19 +159,19 @@ Member types and constants
 
 .. cpp:function:: constraints& constraints::set_numa_id(numa_node_id id)
 
-    Sets the `numa_id` to the provided ``id``. Returns the updated constraints object.
+    Sets the `numa_id` to the provided ``id``. Returns the reference to the updated constraints object.
 
 .. cpp:function:: constraints& constraints::set_max_concurrency(int maximal_concurrency)
 
-    Sets the `max_concurrency` to the provided ``maximal_concurrency``. Returns the updated constraints object.
+    Sets the `max_concurrency` to the provided ``maximal_concurrency``. Returns the reference to the updated constraints object.
 
 .. cpp:function:: constraints& constraints::set_core_type(core_type_id id)
 
-    Sets the `core_type` to the provided ``id``. Returns the updated constraints object.
+    Sets the `core_type` to the provided ``id``. Returns the reference to the updated constraints object.
 
 .. cpp:function:: constraints& constraints::set_max_threads_per_core(int threads_number)
 
-    Sets the `max_threads_per_core` to the provided ``threads_number``. Returns the updated constraints object.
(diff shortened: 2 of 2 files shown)


source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
```diff
@@ -27,11 +27,15 @@ A class that represents an explicit, user-managed task scheduler arena.
                 };
 
                 struct constraints {
-                    numa_node_id numa_node;
-                    int max_concurrency;
-
-                    constraints(numa_node_id numa_node_       = task_arena::automatic,
-                                int          max_concurrency_ = task_arena::automatic);
```
ivankochin, line 34: @alexey-katranov @aleksei-fedotov 

Folks, how do you think, do we want to remove this constructor from the spec?

For me, it seems like the removing makes sense, but it breaks the source backward compatibility.

ivankochin: It was decided that the constructor should exist for all pre-C++20 standards. C++20 standard and later shouldn't have this constructor.

source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
```diff
@@ -123,6 +131,35 @@ Member types and constants
...
 
+    ``core_type`` - An integral logical index uniquely identifying a core type.
+    If set to non-automatic value, then this core type will be considered as preferred for all the
+    threads within the arena.
+
+    .. note::
+
+        core type ID is considered valid if it was obtained through ``tbb::info::core_types()``.
+
+    ``core_type`` - The maximum number of threads that can be scheduled to one core simultaneously.
```
aleksei-fedotov, line 142: A typo:
```suggestion
    ``max_threads_per_core`` - The maximum number of threads that can be scheduled to one core simultaneously.
```

source/elements/oneTBB/source/info_namespace.rst
```diff
@@ -16,8 +16,13 @@ Interfaces to query information about execution environment.
...
     namespace tbb {
         using numa_node_id = /*implementation-defined*/;
+        using core_type_id = /*implementation-defined*/;
+
         namespace info {
             std::vector<numa_node_id> numa_nodes();
+            std::vector<core_type_id> core_types();
+
+            int default_concurrency(task_arena::constraints c);
             int default_concurrency(numa_node_id id = oneapi::tbb::task_arena::automatic);
```
aleksei-fedotov, line 26: As far as I understand, there is no necessity in having `default_concurrency(numa_id)` in presence of a more general `default_concurrency(constraints)` function. Can we remove it from the spec in a similar way as constructor for `constraints` structure below is going to be removed? I.e. `default_concurrency(numa_id)` does not exist in the specification starting from C++20.

ivankochin: I don't think that we really should delete this overload. The constructor is removed in case of using C++20 to allow designated initialization. Removing the overload doesn't bring any new abilities to customers, so I think we should leave it as is. @alexey-katranov what do you think about it?

ivankochin: It was discussed offline that removing the overload is not worth the legacy interfaces break.

aleksei-fedotov (approved):
Looks good to me!

source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
```diff
@@ -113,16 +120,59 @@ Member types and constants
...
+
+        core type ID is considered valid if it was obtained through ``tbb::info::core_types()``.
+
+.. cpp:member:: int constraints::max_threads_per_core
+
+    The maximum number of threads that can be scheduled to one core simultaneously.
+
+.. cpp:function:: constraints::constraints(numa_node_id numa_node_ = task_arena::automatic, int max_concurrency_ = task_arena::automatic)
+
+    Constructs the constriants object with the provided `numa_id` and `max_concurrency` settings.
```
aleksei-fedotov, line 154: A typo:
```suggestion
    Constructs the constraints object with the provided `numa_id` and `max_concurrency` settings.
```

source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
```diff
@@ -113,16 +120,59 @@ Member types and constants
...
+
+    .. note::
+
+        To allow using the C++20 designated initialization the constructor is removed for the C++20 and later standard.
```
aleksei-fedotov, line 158: I don't think we need to explain the reasons here.
Can we just say:
```suggestion
        Starting from C++20 this constructor does not exist.
```
?

aleksei-fedotov (approved):
Looks good to me!

source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
```diff
@@ -113,16 +120,59 @@ Member types and constants
...
+
+    The maximum number of threads that can be scheduled to one core simultaneously.
+
+.. cpp:function:: constraints::constraints(numa_node_id numa_node_ = task_arena::automatic, int max_concurrency_ = task_arena::automatic)
+
+    Constructs the constraints object with the provided `numa_id` and `max_concurrency` settings.
+
+    .. note::
+
+        Starting from C++20 this constructor does not exist.
```
ivankochin, line 158:
```suggestion
        To support designated initialization this constructor is omitted starting from C++20. Aggregate initialization is supposed to be used instead.
```

source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
```diff
@@ -113,16 +120,59 @@ Member types and constants
 
     Represents limitations applied to threads within ``task_arena``.
```
ivankochin, line 122:
```suggestion
    Represents limitations applied to threads within ``task_arena``.

    Starting from C++20 this class should be an aggregate type to support the designated initialization.

```

PR #400, opened 2022-01-06
update oneDNN Graph with latest release

wuxun-zhang:
As title


wuxun-zhang:
@rscohn2 Please help review. thanks. cc @Taolv

PR #408, opened 2022-02-11
[oneMKL][LAPACK] Add missing constraint on parameter 'n' for orgqr functions

rachelertl:
Changes made: Add constraint 'n <= m'

Rationale: The existing specification erroneously omits this constraint


spencerpatty (approved):
These changes look fine to me.  Not a big change, just clarifying the relationship between m and n.  Once it passes validation, I am fine with merging it.

Issue #379, opened 2021-10-01
[RFC][oneMKL][Data Fitting] extend oneMKL Specification with Data Fitting functionality


capatober:
It can be useful for developers to use such optimized math algorithms as
* spline-based interpolation
* derivatives' calculation
* integration

To implement it we can have the Data Fitting component for oneMKL.

PR #413, opened 2022-03-17
[oneMKL] add Data Fitting domain to oneMKL

andreyfe1:
The PR introduces the new (experimental) domain of oneMKL: Data Fitting. It provides spline-based interpolation capabilities that can be used for spline construction (Linear, Cubic, Quadratic etc.), to perform cell-search operations, and to approximate functions, function derivatives, or integrals.

The PR is addressing the https://github.com/oneapi-src/oneAPI-spec/issues/379 issue


source/elements/oneMKL/source/domains/data_fitting/interpolate.rst
```diff
@@ -0,0 +1,106 @@
...
+
+If the sites do not belong to interpolation interval ``[a, b]``, the library uses:
+
+  - interpolant :math:`I_0` coefficients computed for interval :math:`[x_0, x_1)` for the
+    computations at the sites to the left of ``a``.
+  - interpolant :math:`I_{n-2}` coefficients computed for interval
+    :math:`[x_{n-2}, x_{n-1})` for the computations at the sites to the right of ``b``.
+
+Interpolation algorithm depends on interpolant's type (e.g., for cubic spline
+interpoilation evaluation of third-order polynomial is performed to obtain function values).
```
spencerpatty, line 20: spelling check:  interpoilation -> interpolation

source/elements/oneMKL/source/domains/data_fitting/interpolate.rst
```diff
@@ -0,0 +1,106 @@
...
+.. code:: cpp
+
+  #include<oneapi/mkl/experimental/data_fitting.hpp>
+
+Namespace
+---------
+
+.. code:: cpp
+
+  oneapi::mkl::experimental::data_fitiing
```
spencerpatty, line 34: spell check: data_fitting

source/elements/oneMKL/source/domains/data_fitting/terms.rst
```diff
@@ -0,0 +1,324 @@
...
+.. contents::
+    :local:
+    :depth: 2
+
+Glossary
+--------
+
+Assume we need to interpolate a function ``f(x)`` on the ``[a, b)`` interval using splines.
+
+Let's break ``[a, b)`` into sub-intervals by ``n`` points :math:`x_i`
```
spencerpatty, line 15: The language seems a little too informal.  Maybe change "Let's break"  into something like "We break "

source/elements/oneMKL/source/domains/data_fitting/terms.rst
```diff
@@ -0,0 +1,324 @@
...
+called partition points or simply **partition**.
```
spencerpatty, line 16: simply a **partition**  ?
           ^

spencerpatty, line 16: The language seems to be a little informal and not exactly clear to me what is what. Do you mean this:  

We define a **partition** of ``[a, b)`` to be a collection of :math:`n` sub-intervals :math:`\left\{ \left[x_{i-1}, x_{i}\right) \vert i = 1, \dots, n\right\}` where  :math:`x_0 = a`, and :math:`x_n = b`.  The :math:`n+1` points, :math:`x_i`, are called the **partition points**.

source/elements/oneMKL/source/domains/data_fitting/terms.rst
```diff
@@ -0,0 +1,324 @@
...
+**Function values** at these points (:math:`y_i = f(x_i), i=1,\cdots,n`) are also given.
```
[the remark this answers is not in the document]
spencerpatty: same through out the document

source/elements/oneMKL/source/domains/data_fitting/terms.rst
```diff
@@ -0,0 +1,324 @@
...
+.. math::
+  P_i\left( x \right) =
+  c_{1,i} +
+  c_{2,i}\left( x - x_i \right) +
+  c_{3,i}{\left( x - x_i \right)}^2 + \cdots +
+  c_{k-1,i}{\left( x - x_i \right)}^k.
+
+:math:`c_{j,i}, j=1,\cdots,k, i=1,\cdots,n` are called spline **coefficients**.
+
+Function is interpolated at points of [a, b).
```
spencerpatty, line 40: Given a interpolant spline of `:math:`f(x)` on a partition of :math:`[a,b)`, The interpolant function can be evaluated at any point in the interval :math:`[a,b)`.  Such points to evaluate the interpolant at will be referred to as **interpolation sites** or simply **sites**.  Sites can correspond to partition points, but are not required to be.  They must however be in the interval :math:`[a,b)` to be considered interpolation sites.


side question:  can they also be at the value "b" ?  so this could be :math:`[a,b]` ?

source/elements/oneMKL/source/domains/data_fitting/terms.rst
spencerpatty: I guess reading below, your partition is not the subintervals, but the collection of points ?  IS that mathematically correct?  or is the partition really the set of sub-intervals adn the partition points or nodes are the end points of each sub--interval ?

source/elements/oneMKL/source/domains/data_fitting/terms.rst
```diff
@@ -0,0 +1,324 @@
...
+
+       :math:`x_{i+1} - x_i = x_i - x_{i-1}, i=2,\cdots,n-1`
+   * - Quasi-uniform partition
+     - Partition :math:`\left\{ x_i \right\}_{i=1,\cdots,n}`
+       which meets the constraint with a constant ``C`` defined as:
+       :math:`1 \le M / m \le C`,
+
+       where
+
+       :math:`M = max_{i=1,\cdots,n-1} (\Delta_i)`,
```
spencerpatty, line 66: I like to encase words like max.min etc that show up in math mode in \text{} so that it looks a little better:  IT is your preference, but something to consider:

:math:`M = \text{max}_{i=1,\cdots,n-1} (\Delta_i)`,

source/elements/oneMKL/source/domains/data_fitting/terms.rst
```diff
@@ -0,0 +1,324 @@
...
+
+       :math:`m = min_{i=1,\cdots,n-1} (\Delta_i)`,
+
+       :math:`\Delta_i = x_{i+1} - x_i`
```
spencerpatty, line 70: you should put this \Delta_i definition before M and m definitions.  Is `\Delta_i` used anywhere else?  is it really necessary?  If not, it seems simple enough to just put in the `x_{i+1} - x_{i}` in the previous definitions and avoid introducing `\Delta_i` completely.

source/elements/oneMKL/source/domains/data_fitting/terms.rst
```diff
@@ -0,0 +1,324 @@
...
+   * - Vector-valued function of dimension ``p`` being fit
```
spencerpatty, line 71: what do you mean by the word "fit" here ?  It is not clear

source/elements/oneMKL/source/domains/data_fitting/terms.rst
```diff
@@ -0,0 +1,324 @@
...
+     - :math:`f(x) = (f_1(x),\cdots, f_p(x))`.
+   * - A ``k``-order derivative of function ``f(x)`` at point ``t``
+     - :math:`f^{(k)}(t)`.
+   * - Function ``p`` agrees with function ``f`` at the points
```
spencerpatty, line 75: you are using ``p`` to refer to a function here and two lines above you are using it to refer to a dimension.  This is not clear.  Should we just denote the interpolant of ``f``  as  :math:`\mathcal{I}(f)`  or as ``F`` ?

source/elements/oneMKL/source/domains/data_fitting/terms.rst
```diff
@@ -0,0 +1,324 @@
...
+       :math:`\left\{ x_i \right\}_{i=1,\cdots,n}`.
+     - For every point :math:`\zeta` in sequence :math:`\left\{ x_i \right\}_{i=1,\cdots,n}`
+       that occurs ``m`` times, the equality :math:`p^{(i-1)}(\zeta) = f^{(i-1)}(\zeta)`
+       holds for all :math:`i=1,\cdots,m`.
```
spencerpatty, line 79: I don't understand this statement.  Is that a sequence or just a set ?  What does it mean to occur ``m`` times?  it looks like you are saying the derivatives up to m-1 agree at that point ?  Is that correct ?

source/elements/oneMKL/source/domains/data_fitting/terms.rst
```diff
@@ -0,0 +1,324 @@
...
+     - Mathematical Notation
+   * - Partition
+     - :math:`\left\{ x_i \right\}_{i=1,\cdots,n}`,
+       where :math:`a = x1 < x2< \cdots < xn = b`..
+   * - Uniform partition
+     - Partition :math:`\left\{ x_i \right\}_{i=1,\cdots,n}`
+       which meets the following condition:
+
+       :math:`x_{i+1} - x_i = x_i - x_{i-1}, i=2,\cdots,n-1`
+   * - Quasi-uniform partition
```
spencerpatty, line 59: So you are defining Uniform and Quasi-uniform.  Should you also define the other case as Non-uniform ?

source/elements/oneMKL/source/domains/data_fitting/terms.rst
```diff
@@ -0,0 +1,324 @@
...
+  };
+
+Coefficients Hints
+^^^^^^^^^^^^^^^^^^
+
+Let :math:`\left\{ x_i \right\}_{i=1,\cdots,nx}` is a partition,
+:math:`\left\{ f_j(x) \right\}_{j=1,\cdots,ny}` is a vector-valued function.
+Let cubic spline should be constructed. It means that it requires 4 coefficients
+per each interpolation interval and function value.
+Cofficients are stored in the one-dimensional array with ``4 * (nx - 1) * ny`` elements.
```
spencerpatty, line 155: Let :math:`\left\{ x_i \right\}_{i=1,\cdots,nx}` be a partition of :math:`[a, b)` and
:math:`\mathbf{f}(x) = \left\{ f_j(x) \right\}_{j=1,\cdots,ny}` be a vector-valued function to be interpolated with a cubic spline on this partition.

To create a cubic spline, on each sub-interval in the partition, we require 4 coefficients per component :math:`f_j` of :mathbf{f}(x)`.  The coefficients are stored in a one-dimensional array with :math:` 4 (nx -1)(ny)` elements.

source/elements/oneMKL/source/domains/data_fitting/terms.rst
```diff
@@ -0,0 +1,324 @@
...
+Let :math:`\left\{ f_j(x) \right\}_{j=1,\cdots,ny}` is a vector-valued function,
+:math:`\left\{ s_i \right\}_{i=1,\cdots,ns}` are sites, ``d`` is a number of derivatives (including interpolation values) that needs to be calculated.
+So, size of memory to store interpolation results is ``nsite * ny * d`` elements.
+
+6 different layouts are possible:
+
+ - functions-sites-derivatives
+
+   Let :math:`\left\{ r_{j,i,k} \right\}_{j=1,\cdots,ny, i=1,\cdots,nsite, k=1,\cdots,d}`
+   is an interpolation result that corresponds to the ``i-th`` site,
```
spencerpatty, line 210: **be** an interpolation result...

spencerpatty: similar all through document:    "Let xxx be a/an ..." instead of "Let xxx is a/an ..."

source/elements/oneMKL/source/architecture/api_design.inc.rst
```diff
@@ -16,17 +16,18 @@ oneMKL namespaces
...
+``oneapi::mkl``                             oneMKL base namespace, contains general oneMKL data types, objects, exceptions and routines
+``oneapi::mkl::blas``                       Dense linear algebra routines from BLAS and BLAS like extensions. The oneapi::mkl::blas namespace should contain two namespaces column_major and row_major to support both matrix layouts. See :ref:`onemkl_blas`
+``oneapi::mkl::lapack``                     Dense linear algebra routines from LAPACK and LAPACK like extensions. See :ref:`onemkl_lapack`
+``oneapi::mkl::sparse``                     Sparse linear algebra routines from Sparse BLAS and Sparse Solvers. See :ref:`onemkl_sparse_linear_algebra`
+``oneapi::mkl::dft``                        Discrete and fast Fourier transformations. See :ref:`onemkl_dft`
+``oneapi::mkl::rng``                        Random number generator routines. See :ref:`onemkl_rng`
+``oneapi::mkl::vm``                         Vector mathematics routines, e.g. trigonometric, exponential functions acting on elements of a vector. See :ref:`onemkl_vm`
+``oneapi::mkl::experimental::data_fitting`` Data fitting routines, e.g. interpolate. See :ref:`data_fitting`
```
spencerpatty, line 29: e.g. interpolation.

source/elements/oneMKL/source/domains/data_fitting/cubic.rst
```diff
@@ -0,0 +1,90 @@
...
+
+where
+
+.. math::
+  x \in \left[ x_i, x_{i+1} \right),
+
+.. math::
+  i = 1,\cdots , n-1.
+
+There are a lot of different types of cubic splines: Hermite, natural, Akima, Bessel.
```
spencerpatty, line 25: There are many different kinds of cubic splines: ...

source/elements/oneMKL/source/domains/data_fitting/cubic.rst
```diff
@@ -0,0 +1,90 @@
...
+---------
+
+.. code:: cpp
+
+  oneapi::mkl::experimental::data_fitiing
+
+Hermite Spline
+--------------
+
+Coefficients of Hermite spline are calculated using the following formulas:
```
spencerpatty, line 45: The coefficients of a Hermite spline are calculated ...

source/elements/oneMKL/source/domains/data_fitting/interpolate.rst
```diff
@@ -0,0 +1,106 @@
...
+    :depth: 1
+
+Interpolate function performs computations of function and derivatives values at interpolation sites.
+
+If the sites do not belong to interpolation interval ``[a, b]``, the library uses:
+
+  - interpolant :math:`I_0` coefficients computed for interval :math:`[x_0, x_1)` for the
+    computations at the sites to the left of ``a``.
+  - interpolant :math:`I_{n-2}` coefficients computed for interval
+    :math:`[x_{n-2}, x_{n-1})` for the computations at the sites to the right of ``b``.
```
spencerpatty, line 17: You mean to say that you do allow for extrapolation, but it uses the interpolant function defined on the closest sub-interval in the interpolation interval :math:`[a,b)`?

spencerpatty: This seems different than your definition of site in the terms.rst

source/elements/oneMKL/source/domains/data_fitting/interpolate.rst
```diff
@@ -0,0 +1,106 @@
...
+
+Interpolation algorithm depends on interpolant's type (e.g., for cubic spline
```
spencerpatty, line 19: for a cubic spline, the evaluation of a third-order polynomial is performed to obtain the resulting interpolant value.

source/elements/oneMKL/source/domains/data_fitting/interpolate.rst
```diff
@@ -0,0 +1,106 @@
...
+  oneapi::mkl::experimental::data_fitiing
+
+Syntax
+------
+
+.. code:: cpp
+
+  template <typename Interpolant>
+  sycl::event interpolate(
+      Interpolant& interpolant,
```
spencerpatty, line 43: are these missing the queue as argument?

Also, do you fit into non-member or member function cases as seen in https://github.com/oneapi-src/oneAPI-spec/blob/main/source/elements/oneMKL/source/architecture/execution_model.inc.rst#member-functions ?  If you are of member-function approach, you may want to add some language there to include Data fitting in list with DFt and RNG.

source/elements/oneMKL/source/domains/data_fitting/splines.rst
```diff
@@ -0,0 +1,177 @@
...
+   :header-rows: 1
+
+   * - Member function
+     - Description
+   * - ``spline& set_partitions(FpType* input_data, std::int64_t nx, partition_hint PartitionHint = partition_hint::non_uniform);``
+     - Set partition values that are specified by the ``input_data`` memory pointer and ``nx`` partition values.
+       Users can provide ``PartitionHint`` to specify the layout of data.
+       Default layout is ``non_uniform``.
+       If ``uniform`` is specified, ``nx`` must equals to ``2`` and
+       ``input_data`` must contain only 2 values the left and the right borders of partition.
```
spencerpatty, line 101: If ``uniform`` is specified, ``nx`` must be equal to ``2`` and ``input_data`` must contain only 2 values corresponding to the left and the right borders of the sub-interval defined by the partition.

source/elements/oneMKL/source/domains/data_fitting/splines.rst
```diff
@@ -0,0 +1,177 @@
...
+       Otherwise, behavior is undefined.
+       If ``input_data`` layout doesn't satisfy ``PartitionHint``, behavior is undefined.
+       Returns a reference to the spline object for which partitions are set.
+
+       Example, for ``uniform``. Let :math:`\left\{ i \right\}_{i=1,\cdots,n}` is a partition.
+       So, ``input_data`` must contain only 2 values: 1, ``n``.
+   * - ``spline& set_function_values(FpType* input_data, function_hint FunctionHint = storage_hint::row_major);``
+     - Set function values that are specified by the ``input_data`` memory pointer.
+       Number of function values must equals to ``ny * nx`` elements.
```
spencerpatty, line 110: Number of function values must be ``ny * nx``.

source/elements/oneMKL/source/domains/data_fitting/splines.rst
```diff
@@ -0,0 +1,177 @@
...
+       Users can provide ``FunctionHint`` to specify the layout of data.
+       Default layout is ``row_major``.
+       If ``input_data`` layout doesn't satisfy ``FunctionHint``, behavior is undefined.
+       Returns a reference to the spline object for which function values are set.
+   * - ``spline& set_coefficients(FpType* data, coefficient_hint CoeffHint = storage_hint::row_major);``
+     - Set coefficients that are specified by the ``data`` memory pointer.
+       Number of coefficients in the memory must equals to the return value of ``get_required_coeffs_size()``.
+       Users can provide ``CoeffHint`` to specify the layout of data.
+       Default layout is ``row_major``.
```
spencerpatty, line 119: Users can provide a ``CoeffHint`` to specify the layout of data, with the default layout of ``row_major``.

source/elements/oneMKL/source/domains/data_fitting/splines.rst
```diff
@@ -0,0 +1,177 @@
...
+      using value_type = FpType;
+      using spline_type = SplineType;
+
+      spline(
+        const sycl::queue& q,
+        std::int64_t ny = 1);
+
+      spline(
+        const sycl::device& dev,
+        const sycl::context& ctx,
```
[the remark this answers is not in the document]
capatober: No problem. I've been inspired by SYCL 2020 preparing interfaces :)

source/elements/oneMKL/source/domains/data_fitting/terms.rst
capatober: You're right partition means a collection of points in our terms.
Regarding math. It's a collection of points. I'm not sure if it's a good proof (sorry of not), but here it is: https://en.wikipedia.org/wiki/Partition_of_an_interval

What if we say following?
>We define a **partition** of `[a, b)` to be a collection of :math:`n` points
(called **partition points**) :math:`\left\{ x_{i}\right}, i = 1, \dots, n` such as
:math:`a = x_1 < x_2 < \dots < x_n = b`.

source/elements/oneMKL/source/domains/data_fitting/terms.rst
capatober: Actually sites can be `< a` or `>= b`, but in this case first or last sub-interval will be used for interpolation.

source/elements/oneMKL/source/domains/data_fitting/interpolate.rst
capatober: > You mean to say that you do allow for extrapolation, but it uses the interpolant function defined on the closest sub-interval in the interpolation interval :math:`[a,b)`?

Correct.

> This seems different than your definition of site in the terms.rst

Statements here are correct. Will rework "terms.rst"

source/elements/oneMKL/source/domains/data_fitting/terms.rst
capatober: I mean that function values equals and all derivatives equals at the given set of points. Will rephrase

source/elements/oneMKL/source/domains/data_fitting/interpolate.rst
```diff
@@ -0,0 +1,106 @@
...
+      interpolate_hint ResultHint = interpolate_hint::funcs_sites_ders,
+      site_hint SiteHint = site_hint::non_uniform); // (1)
+
+  template <typename Interpolant>
+  sycl::event interpolate(
+      Interpolant& interpolant,
+      typename Interpolant::fp_type* sites,
+      std::int64_t n_sites,
+      typename Interpolant::fp_type* results,
+      std::bitset<32> der_indicator,
```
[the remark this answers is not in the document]
spencerpatty: I suppose so, but if I were you, I would choose a general pattern so it is predictable for all names...   lower snake case with full words ...  I see you sometimes shortening words sometimes not shortening them...  is there a pattern to it?  maybe you wan to rethink all the names into a single pattern :)

See if you can write down a general naming pattern:  recall this spec is supposed to be written so that another developer could implement the library, so you want to describe naming patterns when applicable in case someone wants to suggest an extension.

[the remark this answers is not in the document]
spencerpatty: on the other hand, I am ok with just a few shortened cases as long as it is readable... but you should review all your names and see if there is a general pattern that you can (and want to) write down... you may discover you want to be more consistent, or that you are ok with it :)  To me this is just one aspect of the full review.

source/elements/oneMKL/source/domains/data_fitting/interpolate.rst
spencerpatty: What about changing the spec to say something like this:

> In general, each oneMKL non-member computational routine takes a sycl::queue reference as its first parameter. 
 In some exceptional cases, like in Data Fitting, it may instead take an object as first parameter that itself contains a reference to the sycl execution context.

spencerpatty: why is a queue not passed in?  why is a context sufficient ?  Is there a future call which adds the queue ?   How does work get done ?

source/elements/oneMKL/source/domains/data_fitting/terms.rst
spencerpatty: Yeah, I would say that either way can be fine, but maybe you want to say that in the partition definition something like the following:

>  In general when there is no condition with respect to the distances between the :math:`x_i` , the partition is referred to as a **non-uniform partition**.

source/elements/oneMKL/source/domains/data_fitting/terms.rst
```diff
@@ -49,38 +49,37 @@ Mathematical Notation in the Data Fitting Component
    * - Concept
      - Mathematical Notation
    * - Partition
-     - :math:`\left\{ x_i \right\}_{i=1,\cdots,n}`,
-       where :math:`a = x1 < x2< \cdots < xn = b`..
+     - :math:`\left\{ x_i \right\}_{i=1,\dots,n}`,
+       where :math:`a = x1 < x2< \dots < xn = b`.
```
spencerpatty, line 53: where :math:`a = x_1 < x_2 < \dots < x_n = b`.

source/elements/oneMKL/source/domains/data_fitting/terms.rst
spencerpatty: m is a good dimension, but it is also overloaded in the line before as the minimal (x_{i+1} - x_i) ...  maybe you want to change those to something like max_sub_interval  and min_sub_interval ??

source/elements/oneMKL/source/domains/data_fitting/terms.rst
spencerpatty: that looks better

source/elements/oneMKL/source/domains/data_fitting/terms.rst
spencerpatty: I should have used \dots here as well instead of \cdots

source/elements/oneMKL/source/domains/data_fitting/interpolate.rst
capatober: Actually the queue is inside (See here: https://github.com/oneapi-src/oneAPI-spec/pull/413/files/#diff-2361f70703f03fec4684fd287b54dae6e0b303c2f095bab62be9b5bc536b51f8R34-R41).
Sorry, I might misspell it trying to explain it more generally.

source/elements/oneMKL/source/domains/data_fitting/interpolate.rst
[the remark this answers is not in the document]
capatober: Ok. I see. Yeah, we need to think about general pattern and exceptions for it (if any). As for me interpolation sites will look too long... However, we definitely need to think about it. Thanks

spencerpatty:
Hi @capatober it looks like the ci build is failing in latexpdf step, you may want to try building locally to figure out what is going wrong and fix it

source/elements/oneMKL/source/domains/data_fitting/splines.rst
```diff
@@ -0,0 +1,185 @@
...
+      spline(
+        const sycl::queue& queue,
+        std::int64_t ny = 1,
+        bool were_coeffs_computed = false);
+
+      spline(
+        const sycl::device& device,
+        const sycl::context& context,
+        std::int64_t ny = 1,
+        bool were_coeffs_computed = false);
```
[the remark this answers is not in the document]
capatober: > is this constructor when you need to pass the queue in the interpolate function ?

Correct.

> Running on different queues at the same time with the same internal spline data ? 

It's helpful when users do interpolation in different queues, splitting `sites` between them: the spline object is the same, but `sites` are different.

> Do you have anywhere in this document that explains the rationale behind having these two constructors for the spline?

Yes. It's [here](https://github.com/oneapi-src/oneAPI-spec/pull/413/files/#diff-2361f70703f03fec4684fd287b54dae6e0b303c2f095bab62be9b5bc536b51f8R92-R97). Maybe we need to describe cases clearly here.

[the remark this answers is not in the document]
capatober: Agree... Thanks for pointing it out!

spencerpatty:
Hi @andreyfe1 can you update on the status of this PR?  Do you have a sense of when you want to target getting this in?  We are just passing a provisional 1.2 release now and the next one will be around same time as oneMKL 2023.0 product release (in november or so) What needs to be done more for this ?

andreyfe1:
Hi @spencerpatty!
I think this is not for 1.2. I'm trying to get feedbacks from [potential] users. Also, anyone who interested in these APIs can post their thoughts in this PR

spencerpatty:
@andreyfe1 and @paveldyakov any reason why we shouldn't also merge this PR for Spec 1.3 ?

andreyfe1:
Hi @spencerpatty. As it's mentioned one comment above we are still waiting for feedbacks. That's why we shouldn't also merge this PR for Spec 1.3

PR #420, opened 2022-06-13
Add transpose routines to BLAS-like extensions

andrewtbarker:
This PR adds transpose and batch-transpose BLAS-like extensions `imatcopy`, `omatcopy`, `omatadd`, `imatcopy_batch`, `omatcopy_batch`, and `omatadd_batch` to the spec.


source/elements/oneMKL/source/domains/blas/omatadd_batch.rst
```diff
@@ -0,0 +1,517 @@
...
+      * -  ``std::complex<float>`` 
+      * -  ``std::complex<double>`` 
+
+.. _onemkl_blas_omatadd_batch_buffer:
+
+omatadd_batch (Buffer Version)
+------------------------------
+
+.. rubric:: Description
+
```
aaronjohnson, line 41: Please add a note here like omatadd_batch about which APIs this buffer version supports.

source/elements/oneMKL/source/domains/blas/imatcopy.rst
```diff
@@ -0,0 +1,299 @@
...
+.. rubric:: Description
+
+The ``imatcopy`` routine performs an in-place scaled
+matrix copy or transposition.
+
+The operation is defined as:
+
+.. math::
+
+      AB \leftarrow \alpha * op(AB)
```
aaronjohnson, line 24: I think having the matrix just be `A` would be better.    Unless there is a reason to name the matrix `AB` ?

aaronjohnson (changes requested):
Is there a place where a table summarizing which APIs ( group, strided ) are supported for each of ( Buffer, USM ) ?   Not sure if this is done elsewhere in the spec, but as I was reading I thought it would be helpful to see all this in one place.   Perhaps per each type of API ?
      
              |  group    |     strided
------------------------------
buffer   |      yes      |      no
USM      |     yes      |      yes


what do you think?

source/elements/oneMKL/source/domains/blas/imatcopy.rst
andrewtbarker: The argument in the programming API is named `AB` and I think it makes some sense to have the math description consistent with that. That's also what's done in the [oneMKL developer reference](https://www.intel.com/content/www/us/en/develop/documentation/onemkl-developer-reference-c/top/blas-and-sparse-blas-routines/blas-like-extensions/mkl-imatcopy.html). But I can certainly change it if that's better.

andrewtbarker:
@aaronjohnson I addressed your comments and added the table you suggested, let me know what you think.

source/elements/oneMKL/source/domains/blas/imatcopy.rst
spencerpatty: What is the purpose of using the `AB` anywhere?  I don't understand where the B part of the name comes in.   what is the intuition over @aaronjohnson's suggestion to use a basic `A` or even `X` ?

andrewtbarker: This notation predates any of my work, I'm just trying to be consistent and I'm happy to change it for clarity.

One argument for the current notation is that we have separate `lda` and `ldb` parameters here - what should they be named if `AB` becomes `matrix`? Maybe `input_ld` and `output_ld`? (and similarly for `stride_a` and `stride_b` in the batch API). I think the current naming is aiming for some consistency with `omatcopy`, this routine does the same thing but it just so happens `A` and `B` live in the same memory so we call that memory `AB`. (not saying that's ideal, but I think that's the motivation)

aaronjohnson: And a new person reading the spec, I would find a single letter A, B or X to be less surprising.     I had pause wondering what was special about AB.     Either a single letter OR an explanation for the two would suffice.   Thank you, Andrew.

andrewtbarker: I've redone the `imatcopy` and `imatcopy_batch` descriptions to avoid a matrix `AB`. I'm now denoting the matrix as a mathematical object as `C` and the array/buffer that holds its memory as `matrix_in_out`. I think that distinction makes sense and everything is clearer now but please let me know what you think.

source/elements/oneMKL/source/domains/blas/omatcopy.rst
```diff
@@ -0,0 +1,316 @@
...
+
+.. code-block:: cpp
+
+   namespace oneapi::mkl::blas::column_major {
+       void omatcopy(sycl::queue &queue,
+                     oneapi::mkl::transpose trans,
+                     std::int64_t m,
+                     std::int64_t n,
+                     T alpha,
+                     sycl::buffer<T, 1> &a,
```
spencerpatty, line 64: does it make sense t ochange this to a=matrix_in  and b = matrix_out  with ld_in and ld_out as well ?

source/elements/oneMKL/source/domains/blas/imatcopy_batch.rst
```diff
@@ -0,0 +1,507 @@
...
+.. rubric:: Description
+
+The buffer version of ``imatcopy_batch`` supports only the strided API.
+
+The operation for the strided API is defined as:
+::
+
+   for i = 0 … batch_size – 1
+       C is a matrix at offset i * stride in matrix_array_in_out
+       C = alpha * op(C)
```
spencerpatty, line 67: should we use <-  instead of = here ?

source/elements/oneMKL/source/domains/blas/imatcopy_batch.rst
```diff
@@ -0,0 +1,507 @@
...
+copies or transpositions. They are batched versions of :ref:`onemkl_blas_imatcopy`,
+but the ``imatcopy_batch`` routines perform their operations with
+groups of matrices. Each group contains matrices with the same parameters.
+
+There is a *strided API*, in which the matrices in a batch are a set
+distance away from each other in memory and in which all matrices
+share the same parameters (for example matrix size), and a more
+flexible *group API* where each group of matrices has the same
+parameters but the user may provide multiple groups that have
+different parameters.
```
spencerpatty, line 27: Below it looks like you have a buffer api that is strided, and two USM apis -- strided and group.  Should buffer also ahve the group api?  or is there a reason why you don't provide it?  If there is a reason, it probably makes sense to add that discussion to the spec as well...

andrewtbarker: The group API needs an array of pointers (ie `float `**`) and my understanding is sycl buffers do not handle this, so we do not support the group API with buffers. This is true for all our batched routines, eg [gemm](https://spec.oneapi.io/versions/latest/elements/oneMKL/source/domains/blas/gemm.html#gemm-buffer-version).

We could certainly provide some reasoning for that here, but then would we also add it for the other batched routines? What language do you suggest?

source/elements/oneMKL/source/domains/blas/omatcopy.rst
andrewtbarker: Using `a`, `b`, `lda`, `ldb` is extremely common throughout the BLAS spec so I'm inclined to leave it as is. We changed it in `imatcopy` because there we had some genuine confusion about what the meaning was, here I think it is much clearer. I guess the question is if we want this to be consistent with `imatcopy` or consistent with all the other BLAS and BLAS-like routines? I vote for the latter, leaving `imatcopy` as a special case.

source/elements/oneMKL/source/domains/blas/imatcopy_batch.rst
spencerpatty: I see, would an array of sycl::buffers would be possible ?  I suppose you have discussed things internally within blas team so maybe there are reasons that wouldn't work ? It seems like sycl::buffer * is a meaningful object and  sycl::buffer ** is also reasonable, but it somewhat violates the C++ style of sycl objects,  maybe a std::vector<sycl::buffer> & could be better ??  :)

Barring any additions, it would make sense to say something like 

`The Group API argument structure is more natural when using USM pointers than sycl::buffers, so we only implement it for USM inputs.  On the other hand, the Strided API works for both cases and so USM and sycl::buffer APIs are provided.`

andrewtbarker: @lhuot Can you comment on the group API buffer limitation?

I'll add something like what you propose to the routines added in this PR, but won't touch any of the existing documents.

lhuot: You cannot create vector or array of buffer accessor so that mean if you had such batch group API with buffer input you won't be able to implement real batch optimization with kernel accessing several matrices.

source/elements/oneMKL/source/domains/blas/omatcopy.rst
mmeterel: I vote to be consistent with BLAS domain as well - since these APIs are now included in BLAS headers/documentation.

mmeterel:
@andrewtbarker is this good to merge?

andrewtbarker:
@mmeterel we have a TAB meeting on these changes next week, I think it makes sense to wait until after the meeting to merge.

andrewtbarker:
@aaronjohnson @mmeterel @lhuot @spencerpatty I've modified the spec based on feedback from TAB and others:
1. Added documentation of in-place use of `omatadd` to do `C += alpha A` and other previously unsupported cases;
2. Added `geam` as a simple wrapper around `omatadd`;
3. Added `omatcopy2` because it was missing from spec;
4. More uniform references to strided-batch and group-batch for all the APIs.

I'd appreciate any comments or suggestions.

andrewtbarker:
In my view this is ready to merge - let me know if you have any concerns.

source/elements/oneMKL/source/domains/blas/imatcopy.rst
```diff
@@ -0,0 +1,299 @@
...
+..
+.. SPDX-License-Identifier: CC-BY-4.0
+
+.. _onemkl_blas_imatcopy:
+
+imatcopy
+========
+
+Computes an in-place scaled matrix transpose or copy operation
+using a general matrix.
```
spencerpatty, line 11: maybe add the word dense,  as in:    a general dense matrix  ?

source/elements/oneMKL/source/domains/blas/imatcopy_batch.rst
```diff
@@ -0,0 +1,510 @@
...
+groups of matrices. Each group contains matrices with the same parameters.
+
+There is a *strided API*, in which the matrices in a batch are a set
+distance away from each other in memory and in which all matrices
+share the same parameters (for example matrix size), and a more
+flexible *group API* where each group of matrices has the same
+parameters but the user may provide multiple groups that have
+different parameters. The group API argument structure is better
+suited to USM pointers than to ``sycl::buffer`` arguments, so we
+only implement it for USM inputs. The strided API works with both USM
```
spencerpatty, line 29: a better word would be "only specify it for USM inputs" ...

spencerpatty (approved):
Discussed a few concerns offline (mainly omatcopy2's use case and justification for being in the specification is a little light, and a recommendation to clarify the op(X) breadth of support to be identity op (no-transpose), real transpose, complex hermitian, complex transpose (no complex conjugate), and then clarify if complex conjugate no transpose should also be supported)  

Generally looks good and approval is merited

andrewtbarker:
@spencerpatty I made most of the changes you suggested (add "dense" and use "specify" instead of "implement") in several places. Regarding transpose options: I clarified the transpose datatype description [here](https://github.com/oneapi-src/oneAPI-spec/pull/420/files#diff-5202a2c59db032e3de63e3cf1f485d9001284146663a60547e1c11ee5510b4a5), which is referenced multiple times from the spec, but I left the individual pages because their description of the transpose parameters is consistent with elsewhere in the spec (ie `blas::gemm`).

PR #446, opened 2022-10-31
OneVPL v2.8

tletnes:


jonrecker (approved):
Thanks. Looks good to me. I also tried building the spec locally from your branch, and that worked fine.

PR #447, opened 2022-11-02
oneDPL spec: Data placement support for parallel algorithms

MikeDvorskiy:
oneDPL spec: Data placement support for parallel algorithms.


akukanov:
These changes were taken into https://github.com/oneapi-src/oneAPI-spec/pull/450 and reworked.

Issue #462, opened 2023-01-06
Issue on page /elements/oneMKL/source/domains/blas/syrk.html


jakub-homola:
`oneapi::mkl::blas::syrk` - input parameters - the description of the `upper_lower` parameter states

> Specifies whether `A`’s data is stored in its upper or lower triangle. ...

but it should probably be 

> Specifies whether `C`’s data ...

this appears both at the buffer and the USM version.

the version of the docs I am seeing is 1.2-rev-1

mmeterel:
@jakub-homola The fix is merged. Thanks again for catching. Closing this issue. Please re-open if needed.

PR #465, opened 2023-01-12
[oneMKL][BLAS] Fix syr*k typo

mmeterel:
This PR is closing #462


mmeterel:
@andrewtbarker Could you please take a look at this?

andrewtbarker (approved):
This looks right to me. I wish we could somehow use the same source for the DPCPP docs and in the spec?

mmeterel:
@rscohn2 I see the CI checks are failing. I don't think it is related to this PR. Do you have any idea?

mmeterel:
> This looks right to me. I wish we could somehow use the same source for the DPCPP docs and in the spec?

Thanks for checking. Agreed, that has been my hope since we switched to github for DPCPP docs, but I believe there are some format differences and list of exceptions (spec has, docs don't). I don't think these are showstoppers though.

mmeterel:
> @rscohn2 I see the CI checks are failing. I don't think it is related to this PR. Do you have any idea?

@rscohn2 ping

rscohn2:
Can you sync your fork? I believe it will clear the checker error. Go here:
https://github.com/alexandraepan/oneAPI-spec
There should be a Sync fork drop-down below the green <> Code button

PR #468, opened 2023-01-18
Fix ILP32 data model detection with GCC on 32bit ARM

StefanBruens:
Unfortunately, GCC does not define `__ILP32__` on 32 bit ARM (contrary to e.g. x86), so rely on `__arm__` instead.

This does not change API/ABI, but fixes a compilation error for one specific architecture and compiler combination.

Avoids broken workarounds in every downstream user of mfxdefs.h,  e.g.:  
https://github.com/oneapi-src/oneVPL/commit/6182082a12#diff-1e7de1ae2d05   https://github.com/intel/cartwheel-ffmpeg/issues/215

### DIFF
--- a/source/elements/oneVPL/include/vpl/mfxdefs.h
+++ b/source/elements/oneVPL/include/vpl/mfxdefs.h
@@ -68,7 +68,7 @@ extern "C"
     #define MFX_PACK_BEGIN_STRUCT_W_PTR()    MFX_PACK_BEGIN_X(4)
     #define MFX_PACK_BEGIN_STRUCT_W_L_TYPE() MFX_PACK_BEGIN_X(8)
 /* 32-bit ILP32 data model Linux* */
-#elif defined(__ILP32__)
+#elif defined(__ILP32__) || defined(__arm__)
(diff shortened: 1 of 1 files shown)


rscohn2:
@tletnes: Should this change go to the upstream repo?
https://github.com/oneapi-src/onevpl

rscohn2:
Tests will pass if you sync your fork with upstream, but I believe this should go to vpl upstream repo instead.

jonrecker:
In the current model, this repo (oneAPI-spec) is considered the upstream, and oneapi-src/onevpl takes its read-only copies of the headers from oneAPI-spec. (We may wish to change this, however). Because this PR does not modify the API itself but only enables building for an additional architecture+compiler, I'm in favor of making this change now rather than waiting for the next API version update. Adding @varistar for additional review.

mgonchar (approved):
LGTM

tletnes (approved):

tletnes:
@rscohn2 We want to merge this PR It looks like there was an infrastructure issue with the CI checks. I don't see any option to re-run. Can you help re-run the checks?

Issue #470, opened 2023-02-06
Issue on page /elements/oneMKL/source/domains/lapack/lapack.html


jakub-homola:
Hi,

Are all the matrices in the LAPACK domain of oneMKL assumed to be row-major or column-major?

It seems they are assumed to be column-major, but I could not find that information anywhere on the webpage, nor there are any `oneapi::mkl::layout` parameters to the functions. Right now I am mainly concerned about `potrf` and `trtrs`, but knowing that information for the whole LAPACK would be great.

Could you please redirect me to where I can find that information, or add it to the spec/docs?

The version of the spec I am seeing is 1.2-rev-1

mmeterel:
Assigning to @sknepper

sknepper:
Hi @jakub-homola, thanks for the question! The matrices in the LAPACK domain are assumed to be column-major. You're right that we need to add this information to the oneMKL specification. Thanks for finding this!

jakub-homola:
Thanks for the reply.
However, when thinking about it, I think it would be better to add an additional `oneapi::mkl::layout` argument, rather than accepting only column-major matrices. Would it be possible, with respect to backwards compatibility?

sknepper:
Hi @jakub-homola,
When we introduced row major support for the BLAS domain, that was one consideration - to take a parameter that indicated the layout of the matrices (like in CBLAS/LAPACKE). However, we ultimately opted to create separate namespaces as typically an application would use just one layout for the whole application. With namespaces, you don't need to pass an extra layout parameter for every function.

sknepper:
@jakub-homola Thanks again for catching this! I appreciate your interest and thoughtful replies. As we've updated the specification to clarify this, I'd like to ask @mmeterel to close this issue. Please re-open if needed.

And just a note about Intel oneMKL documentation since that was mentioned in the initial report: We note that "Currently, LAPACK DPC++ APIs do not support matrices stored using row major layout." in the Intel® oneAPI Math Kernel Library (oneMKL) - Data Parallel C++ Developer Reference ([link](https://www.intel.com/content/www/us/en/develop/documentation/oneapi-mkl-dpcpp-developer-reference/top/intro-to-onemkl-blas-and-lapack-with-dpcpp.html)). However, we will make this more clear in a future update.  Thanks again!

PR #471, opened 2023-02-07
[oneMKL][LAPACK] Clarify that LAPACK routines require column major storage

sknepper:
This addresses #470 by specifying that LAPACK routines require column major storage. The clarification is added to both the main LAPACK page (https://spec.oneapi.io/versions/1.2-rev-1/elements/oneMKL/source/domains/lapack/lapack.html) and the matrix storage page (https://spec.oneapi.io/versions/1.2-rev-1/elements/oneMKL/source/domains/matrix-storage.html).
Note that the links are to the current 1.2 revision of the specification.

### DIFF
--- a/source/elements/oneMKL/source/domains/lapack/lapack.rst
+++ b/source/elements/oneMKL/source/domains/lapack/lapack.rst
@@ -8,7 +8,7 @@ LAPACK Routines
 +++++++++++++++
 
 oneMKL provides a DPC++ interface to select routines from the Linear Algebra PACKage (LAPACK), as well as several LAPACK-like extension routines.
-LAPACK routines support and require column major layout of matrices.
+LAPACK routines require column major layout of matrices.
 
 .. include:: lapack-linear-equation-routines.inc.rst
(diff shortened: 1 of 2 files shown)


source/elements/oneMKL/source/domains/lapack/lapack.rst
```diff
@@ -8,6 +8,7 @@ LAPACK Routines
 +++++++++++++++
 
 oneMKL provides a DPC++ interface to select routines from the Linear Algebra PACKage (LAPACK), as well as several LAPACK-like extension routines.
+LAPACK routines support and require column major layout of matrices.
```
rachelertl, line 11: Perhaps saying "require" is enough, since if that layout is required then it must also be supported? But either way gets the message across.

rachelertl (approved):

source/elements/oneMKL/source/domains/lapack/lapack.rst
sknepper: Thanks, @rachelertl , I agree. I updated the notes to only say require.

PR #498, opened 2023-09-08
[oneMKL][spblas] add missing 'const' keyword for alpha, beta, and input array, and remove 'y' from input list in sparse::trsv() spec.

baeseung-intel:
**Goal:**
In this PR, I'd like to update some missing `const` keyword for constant parameters within the spec for `spblas` domain in order to achieve consistency in the spec, and fix some minor information.

**Details:**

- add some missing `const` keyword for constant parameters (e.g. `alpha` and `beta`) in the APIs.
- add `const` keyword for input vector/matrix in some spblas APIs.
- remove `y` vector from the input list of `sparse::trsv()` API spec, since `y` is not an input for `sparse::trsv()` API.


spencerpatty (approved):
These look great!  Approved

PR #500, opened 2023-09-09
[oneDPL] Clarify the version of the C++ standard for the oneDPL spec

akukanov:
Explicitly say that the base C++ standard version is C++17, but the parallel API also implements algorithms & policies from C++20


source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -16,7 +17,7 @@ C++ Standard aligned execution policies
 +++++++++++++++++++++++++++++++++++++++
 
 oneDPL has the set of execution policies and related utilities that are semantically aligned
-with the C++ Standard:
+with the `C++ Standard`_, 6th edition (C++20):
```
rarutyun, line 20:
```suggestion
with the `C++ Standard`_ and 6th edition (C++20):
```
?

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -5,7 +5,8 @@
 Parallel API
 ------------
 
-oneDPL provides the set of algorithms with execution policies as defined by the `C++ Standard`_.
+oneDPL provides the set of parallel algorithms as defined by the `C++ Standard`_,
+including parallel algorithms added in the 6th edition known as C++20.
```
rarutyun, line 9: Do we want to say
```suggestion
including parallel algorithms added in the International Standard ISO/IEC 14882:2020 known as C++20.
```
?

akukanov: According to https://www.iso.org/standard/79358.html, C++20 is the 6th edition of the standard. Since a reference to the standard is already given in this sentence, I think it's quite clear what "the 6th edition" refers to.

I can do this change here, but then the other mentioning of the 6th edition below will look less clear.

timmiesmith (commented):
PR looks good to me. I've provided comments on Ruslan's suggestions.

PR #503, opened 2023-09-19
[oneMKL][BLAS] value_or_pointer wrapper for BLAS USM scalar parameters

andrewtbarker:
Allow either values or pointers for scalar parameters to BLAS functions with USM interfaces.

This would allow users to provide values in device memory for these scalar parameters.

### DIFF
--- a/source/elements/oneMKL/source/domains/value_or_pointer.rst
+++ b/source/elements/oneMKL/source/domains/value_or_pointer.rst
@@ -1,0 +1,106 @@
+.. SPDX-FileCopyrightText: 2019-2020 Intel Corporation
+..
+.. SPDX-License-Identifier: CC-BY-4.0
+
+.. _value_or_pointer:
+
+Scalar Arguments in BLAS
+========================
+
+.. container::
+
+   The USM version of oneMKL BLAS routines for DPC++ will accept either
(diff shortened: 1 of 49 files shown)


source/elements/oneMKL/source/domains/value_or_pointer.rst
```diff
@@ -0,0 +1,106 @@
...
+
+      .. rubric:: Wrapper type
+         :name: wrapper-time
+         :class: sectiontitle
+
+      The USM version of oneMKL BLAS routines use a templated
+      ``value_or_pointer<T>`` wrapper to enable either pointers or values
+      to be passed to routines that take a scalar parameter. This type is in
+      the ``oneapi::mkl::`` namespace and defined in the header file
+      ``oneapi/mkl/types.hpp``.
```
mmeterel, line 67: This seems to be too specific to oneMKL product. For spec, should we consider removing this or make it more general?

mmeterel (approved):
Thanks for the PR. Overall, looks good. I added one question.

source/elements/oneMKL/source/domains/value_or_pointer.rst
andrewtbarker: Agreed, I've removed that sentence.

Issue #508, opened 2023-11-01
[RFC][MKL][DFT] Support external workspaces


hjabird:
The oneMKL DFT interface should support external worspaces as requested in https://github.com/oneapi-src/oneMKL/issues/396.

This feature is present in:
* oneMKL (closed-source)
* cuFFT
* rocFFT

PR with suggested additions to the specification to follow.

hjabird:
Specification for external workspaces was merged.

PR #509, opened 2023-11-01
[oneMKL][DFT] Specification for external workspaces for DFTs

hjabird:
This PR is paired with https://github.com/oneapi-src/oneAPI-spec/issues/508, and introduces external workspaces to the oneMKL DFT specification.

* Add support for external workspaces
* Add the following config_params:
  * WORKSPACE_PLACEMENT
  * WORKSPACE_EXTERNAL_BYTES_REQUIRED
* Add descriptor::set_workspace
* Describe typical usage of an external workspace

### DIFF
--- a/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
+++ b/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
@@ -76,7 +76,7 @@ This is shown in the following example code:
    desc.commit(myQueue);
    // 3. Query the required workspace size
    std::int64_t workspaceBytes{0};
-   desc.get_value(mkl::dft::config_param::WORKSPACE_EXTERNAL_BYTES_REQUIRED, &workspaceBytes);
+   desc.get_value(mkl::dft::config_param::WORKSPACE_EXTERNAL_BYTES, &workspaceBytes);
    // Obtain a sufficiently large USM allocation or buffer. For this example, a USM allocation is used.
    float* workspaceUsm = sycl::malloc_device<float>(workspaceBytes / sizeof(float), myQueue);
    // 4. Set the workspace
--- a/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
+++ b/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
@@ -57,7 +57,7 @@ Typical usage of ``WORKSPACE_EXTERNAL``
 
 Usage of ``config_value::WORKSPACE_EXTERNAL`` typically involves the following order of operations:
 
-#. ``config_value::WORKSPACE_EXTERNAL`` is set for the uncommitted descriptor.
+#. ``config_value::WORKSPACE_EXTERNAL`` is set for the uncommitted descriptor's ``config_param::WORKSPACE_EXTERNAL``.
 #. The descriptor is committed.
 #. The required workspace size is queried.
 #. A workspace of sufficient size is provided to the descriptor.
--- a/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
+++ b/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
@@ -57,6 +57,7 @@ Usage of ``WORKSPACE_EXTERNAL`` typically involves the following order of operat
 #. The required workspace size is queried.
 #. A workspace of sufficient size is provided to the descriptor.
 #. Compute functions following the type of external workspace provided are called.
+#. The user is responsible for freeing the external workspace.
 
 This is shown in the following example code:
 
--- a/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
+++ b/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
@@ -9,15 +9,17 @@ Workspace placement
 
 DFT implementations often require temporary storage for intermediate data whilst computing DFTs.
 This temporary storage is referred to as a *workspace*.
-Whilst this is managed automatically by default (``config_param::WORKSPACE_AUTOMATIC``), 
-it may be preferable to provide an external workspace (``config_param::WORKSPACE_EXTERNAL``) for the following reasons:
+Whilst this is managed automatically by default (``config_value::WORKSPACE_AUTOMATIC`` 
+set for ``config_param::WORKSPACE_PLACEMENT``), it may be preferable to provide an external 
+workspace (``config_value::WORKSPACE_EXTERNAL`` set for ``config_param::WORKSPACE_PLACEMENT``) 
+for the following reasons:
 
-* To reduce the number of GPU mallocs / frees
-* To reduce memory consumption
+* to reduce the number of GPU mallocs / frees;
+* to reduce memory consumption.
 
-For some backends and configurations, ``config_param::WORKSPACE_EXTERNAL`` may reduce performance.
+For some backends and configurations, ``config_value::WORKSPACE_EXTERNAL`` may reduce performance.
 
-A typical workflow for using ``config_param::WORKSPACE_EXTERNAL`` is given in the section :ref:`onemkl_dft_typical_usage_of_workspace_external`.
+A typical workflow for using ``config_value::WORKSPACE_EXTERNAL`` is given in the section :ref:`onemkl_dft_typical_usage_of_workspace_external`.
 
 WORKSPACE_PLACEMENT
 +++++++++++++++++++
@@ -43,8 +45,9 @@ When set to ``config_value::WORKSPACE_AUTOMATIC`` the user does not need to prov
 The configuration ``config_param::WORKSPACE_PLACEMENT`` can be set to 
 ``config_value::WORKSPACE_EXTERNAL`` to allow the workspace to be set manually. 
 
-When a descriptor is committed with ``config_value::WORKSPACE_EXTERNAL`` set, 
-the user must provide an external workspace. 
+When a descriptor is committed with ``config_value::WORKSPACE_EXTERNAL`` set 
+for ``config_param::WORKSPACE_PLACEMENT``, the user must provide an external 
+workspace before calling any compute function.
 See :ref:`onemkl_dft_descriptor_set_workspace` and :ref:`onemkl_dft_typical_usage_of_workspace_external`.
 
 .. _onemkl_dft_typical_usage_of_workspace_external:
@@ -52,9 +55,9 @@ See :ref:`onemkl_dft_descriptor_set_workspace` and :ref:`onemkl_dft_typical_usag
 Typical usage of ``WORKSPACE_EXTERNAL``
 +++++++++++++++++++++++++++++++++++++++
 
-Usage of ``WORKSPACE_EXTERNAL`` typically involves the following order of operations:
+Usage of ``config_value::WORKSPACE_EXTERNAL`` typically involves the following order of operations:
 
-#. ``WORKSPACE_EXTERNAL`` is set for the uncommitted descriptor.
+#. ``config_value::WORKSPACE_EXTERNAL`` is set for the uncommitted descriptor.
 #. The descriptor is committed.
 #. The required workspace size is queried.
 #. A workspace of sufficient size is provided to the descriptor.
--- a/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
+++ b/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
@@ -14,7 +14,7 @@ set for ``config_param::WORKSPACE_PLACEMENT``), it may be preferable to provide 
 workspace (``config_value::WORKSPACE_EXTERNAL`` set for ``config_param::WORKSPACE_PLACEMENT``) 
 for the following reasons:
 
-* to reduce the number of GPU mallocs / frees;
+* to reduce the number of mallocs / frees;
 * to reduce memory consumption.
 
 For some backends and configurations, ``config_value::WORKSPACE_EXTERNAL`` may reduce performance.
--- a/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
+++ b/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
@@ -19,7 +19,7 @@ for the following reasons:
 
 For some backends and configurations, ``config_value::WORKSPACE_EXTERNAL`` may reduce performance.
 
-A typical workflow for using ``config_value::WORKSPACE_EXTERNAL`` is given in the section :ref:`onemkl_dft_typical_usage_of_workspace_external`.
+A typical workflow for using ``config_value::WORKSPACE_EXTERNAL`` is given in the section :ref:`below<onemkl_dft_typical_usage_of_workspace_external>`.
 
 WORKSPACE_PLACEMENT
 +++++++++++++++++++
--- a/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
+++ b/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
@@ -1,0 +1,85 @@
+.. SPDX-FileCopyrightText: Codeplay Software
+..
+.. SPDX-License-Identifier: CC-BY-4.0
+
+.. _onemkl_dft_config_workspace_placement:
+
+Workspace placement
+--------------------------------------
+
+DFT implementations often require temporary storage for intermediate data whilst computing DFTs.
+This temporary storage is referred to as a *workspace*.
+Whilst this is managed automatically by default (``config_param::WORKSPACE_AUTOMATIC``), 
+it may be preferable to provide an external workspace (``config_param::WORKSPACE_EXTERNAL``) for the following reasons:
+
+* To reduce the number of GPU mallocs / frees
+* To reduce memory consumption
+
+A typical workflow for using ``config_param::WORKSPACE_EXTERNAL`` is given in the section :ref:`onemkl_dft_typical_usage_of_workspace_external`.
+
+WORKSPACE_PLACEMENT
++++++++++++++++++++
+
+For ``config_param::WORKSPACE_PLACEMENT``, valid configuration values are ``config_value::WORKSPACE_AUTOMATIC`` and ``config_value::WORKSPACE_EXTERNAL``.
+
+.. container:: section
+
+  .. _onemkl_dft_config_value_workspace_automatic:
+
+  .. rubric:: WORKSPACE_AUTOMATIC
+
+The default value for the ``config_param::WORKSPACE_PLACEMENT`` is ``config_value::WORKSPACE_AUTOMATIC``. 
+
+When set to ``config_value::WORKSPACE_AUTOMATIC`` the user does not need to provide an external workspace. The workspace will be automatically managed by the backend library.
+
+.. container:: section
+
+  .. _onemkl_dft_config_value_workspace_external:
+
+  .. rubric:: WORKSPACE_EXTERNAL
+
+The configuration ``config_param::WORKSPACE_PLACEMENT`` can be set to 
+``config_value::WORKSPACE_EXTERNAL`` to allow the workspace to be set manually. 
+
+When a descriptor is committed with ``config_value::WORKSPACE_EXTERNAL`` set, 
+the user must provide an external workspace. 
+See :ref:`onemkl_dft_descriptor_set_workspace` and :ref:`onemkl_dft_typical_usage_of_workspace_external`.
+
+.. _onemkl_dft_typical_usage_of_workspace_external:
+
+Typical usage of ``WORKSPACE_EXTERNAL``
++++++++++++++++++++++++++++++++++++++++
+
+Usage of ``WORKSPACE_EXTERNAL`` typically involves the following order of operations:
+
+#. ``WORKSPACE_EXTERNAL`` is set for the uncommitted descriptor.
+#. The descriptor is committed.
+#. The required workspace size is queried.
+#. A workspace of sufficient size is provided to the descriptor.
+#. Compute functions following the type of external workspace provided are called.
+
+This is shown in the following example code:
+
+.. code-block:: cpp
+
+   // Create a descriptor
+   mkl::dft::descriptor<mkl::dft::precision::SINGLE, dom> desc(n);
+   // 1. Set the workspace placement to WORKSPACE_EXTERNAL
+   desc.set_value(mkl::dft::config_param::WORKSPACE_PLACEMENT, 
+                  mkl::dft::config_value::WORKSPACE_EXTERNAL);
+   // Set further configuration parameters
+   // ...
+   // 2. Commit the descriptor
+   desc.commit(myQueue);
+   // 3. Query the required workspace size
+   std::int64_t workspaceBytes{0};
+   desc.get_value(mkl::dft::config_param::WORKSPACE_EXTERNAL_BYTES_REQUIRED, &workspaceBytes);
+   // Obtain a sufficiently large USM allocation or buffer. For this example, a USM allocation is used.
+   float* workspaceUsm = sycl::malloc_device<float>(workspaceBytes / sizeof(float), myQueue);
+   // 4. Set the workspace
+   desc.set_workspace(workspaceUsm);
+   // 5. Now USM compute functions can be called.
+
+
+**Parent topic:** :ref:`onemkl_dft_enums`
+
--- a/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
+++ b/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
@@ -26,21 +26,17 @@ WORKSPACE_PLACEMENT
 
 For ``config_param::WORKSPACE_PLACEMENT``, valid configuration values are ``config_value::WORKSPACE_AUTOMATIC`` and ``config_value::WORKSPACE_EXTERNAL``.
 
-.. container:: section
+.. _onemkl_dft_config_value_workspace_automatic:
 
-  .. _onemkl_dft_config_value_workspace_automatic:
-
-  .. rubric:: WORKSPACE_AUTOMATIC
+.. rubric:: WORKSPACE_AUTOMATIC
 
 The default value for the ``config_param::WORKSPACE_PLACEMENT`` is ``config_value::WORKSPACE_AUTOMATIC``. 
 
 When set to ``config_value::WORKSPACE_AUTOMATIC`` the user does not need to provide an external workspace. The workspace will be automatically managed by the backend library.
 
-.. container:: section
+.. _onemkl_dft_config_value_workspace_external:
 
-  .. _onemkl_dft_config_value_workspace_external:
-
-  .. rubric:: WORKSPACE_EXTERNAL
+.. rubric:: WORKSPACE_EXTERNAL
 
 The configuration ``config_param::WORKSPACE_PLACEMENT`` can be set to 
 ``config_value::WORKSPACE_EXTERNAL`` to allow the workspace to be set manually. 
--- a/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
+++ b/source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
@@ -15,6 +15,8 @@ it may be preferable to provide an external workspace (``config_param::WORKSPACE
 * To reduce the number of GPU mallocs / frees
 * To reduce memory consumption
 
+For some backends and configurations, ``config_param::WORKSPACE_EXTERNAL`` may reduce performance.
+
 A typical workflow for using ``config_param::WORKSPACE_EXTERNAL`` is given in the section :ref:`onemkl_dft_typical_usage_of_workspace_external`.
 
 WORKSPACE_PLACEMENT
--- a/source/elements/oneMKL/source/domains/dft/descriptor.rst
+++ b/source/elements/oneMKL/source/domains/dft/descriptor.rst
@@ -448,7 +448,7 @@ allocated workspace must only be used with USM compute calls.
 Failing to do this will result in an invalid descriptor for compute calls.
 
 If the workspace is a USM allocation, the user must not use it for other purposes
-in parallel whilst the DFT `compute_forward` or `compute_backward` are in progress.
+in parallel whilst the DFT ``compute_forward`` or ``compute_backward`` are in progress.
 
 This function can be called on committed descriptors where the workspace placement
 is not ``config_value::WORKSPACE_EXTERNAL``. The provided workspace may or may not
@@ -489,7 +489,7 @@ be used in compute calls. However, the aforementioned restrictions will still ap
 
    .. rubric:: Throws
 
-   The `descriptor::set_workspace()` routine shall throw the following exceptions if the associated condition is detected. An implementation may throw additional implementation-specific exception(s) in case of error conditions not covered here:
+   The ``descriptor::set_workspace()`` routine shall throw the following exceptions if the associated condition is detected. An implementation may throw additional implementation-specific exception(s) in case of error conditions not covered here:
    
    :ref:`oneapi::mkl::invalid_argument()<onemkl_exception_invalid_argument>`
       If the provided buffer ``workspaceBuf`` is not sufficiently large, or if the provided USM allocation ``workspaceUSM`` is ``nullptr`` when an external workspace of size greater than zero is required.
--- a/source/elements/oneMKL/source/domains/dft/descriptor.rst
+++ b/source/elements/oneMKL/source/domains/dft/descriptor.rst
@@ -133,7 +133,7 @@ The ``descriptor`` class is defined in the ``oneapi::mkl::dft`` namespace.
      -     Queries the configuration value associated with a particular
            configuration parameter.
    * -     :ref:`onemkl_dft_descriptor_set_workspace`
-     -     Set the external workspace to use when ``config_param::WORKSPACE_PLACEMENT`` is ``config_value::WORKSPACE_EXTERNAL``.
+     -     Sets the external workspace to use when ``config_param::WORKSPACE_PLACEMENT`` is set to ``config_value::WORKSPACE_EXTERNAL``.
    * -     :ref:`onemkl_dft_descriptor_commit`
      -     Commits the ``descriptor`` object to enqueue the operations relevant
            to the (batched) DFT(s) it determines to a given, user-provided
--- a/source/elements/oneMKL/source/domains/dft/descriptor.rst
+++ b/source/elements/oneMKL/source/domains/dft/descriptor.rst
@@ -427,7 +427,7 @@ type ``oneapi::mkl::dft::domain``, ``oneapi::mkl::dft::precision``,
 set_workspace
 +++++++++++++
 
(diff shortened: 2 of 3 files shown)


hjabird:
Note that this PR does not currently include any feature for estimating the required workspace size.

The usecase for this feature is not completely clear, and cannot be supported by the *rocFFT* backend in oneMKL interface library. However, it may nonetheless be required.

source/elements/oneMKL/source/domains/dft/descriptor.rst
```diff
@@ -346,7 +351,84 @@ corresponding configuration values, see :ref:`onemkl_dft_enum_config_param`.
...
+Failing to do this will result in an invalid descriptor for compute calls.
+
+If the workspace provided is a USM allocation shared amongst multiple kernels,
+it must be ensured that the kernels sharing this workspace do not use it simultaneously.
+
+This function can be called on committed descriptors where the workspace placement
+is not ``config_value::WORKSPACE_EXTERNAL``. The provided workspace may or may not
+be used in compute calls. However, the aforementioned restrictions will still apply.
+
+Calling this function multiple times on a committed descriptor is undefined behavior.
```
Rbiessy, line 385: Why do we need this restriction? Is there any harm in allowing to set the workspace multiple times or to switch the workspace between compute functions?

source/elements/oneMKL/source/domains/dft/descriptor.rst
```diff
@@ -346,7 +351,84 @@ corresponding configuration values, see :ref:`onemkl_dft_enum_config_param`.
...
++++++++++++++
+
+Set the workspace for when ``config_param::WORKSPACE_PLACEMENT`` is set to ``config_value::WORKSPACE_EXTERNAL``.
+
+.. rubric:: Description
+
+This function sets the workspace to use when computing DFTs for when an
+external workspace is set. 
+This function may only be called after the descriptor has been committed.
+The provided workspace must be equal to or larger than the required
```
Rbiessy, line 366:
```suggestion
The size of the provided workspace must be equal to or larger than the required 
```

source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
```diff
@@ -0,0 +1,85 @@
...
+Typical usage of ``WORKSPACE_EXTERNAL``
++++++++++++++++++++++++++++++++++++++++
+
+Usage of ``WORKSPACE_EXTERNAL`` typically involves the following order of operations:
+
+#. ``WORKSPACE_EXTERNAL`` is set for the uncommitted descriptor.
+#. The descriptor is committed.
+#. The required workspace size is queried.
+#. A workspace of sufficient size is provided to the descriptor.
+#. Compute functions following the type of external workspace provided are called.
```
Rbiessy, line 59: Should we add a point to mention that the user is responsible for freeing the external workspace?

Rbiessy (commented):
LGTM overall

source/elements/oneMKL/source/domains/dft/descriptor.rst
hjabird: My understanding is that their probably isn't any harm, but I don't know how things work under-the-hood for closed-source FFT libraries. For example a valid optimization might be to only write twiddle factors to device memory once, rather than on every compute call. It also avoid having to discuss edge cases where I suspect this wouldn't be valid.

Rbiessy: Oh right, I was wondering whether the DFT domain could benefit from keeping the workspace reserved. We had similar discussions about the external workspace for the sparse blas domain and the conclusion is that we should allow users to re-use the external workspace as they wish between computes if that means losing some performance. I think it would make sense to do the same for the DFT domain and document that `WORKSPACE_EXTERNAL` can prevent some optimizations because of this. I would be curious to hear the opinion of Helen or Raphael on this.
Do you know if cuFFT and rocFFT mention anything about the restrictions on the external workspace?

source/elements/oneMKL/source/domains/dft/descriptor.rst
hjabird: Done in [a8cbe88](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/a8cbe883d58d2f99e6233c6d49701bb36b297de6).

source/elements/oneMKL/source/domains/dft/descriptor.rst
Rbiessy: By reserved I mean do we allow users to write to the external workspace between 2 compute functions. Whichever the decision we should clarify whether this is allowed. As you said currently with the current wording users could either re-use the workspace for other descriptor or other kernels. I think that makes sense but in that case I don't see the point of having `set_workspace` undefined behavior.
I assume that cuFFT also allows to write on the external workspace between 2 computes if they don't mention any restriction.

hjabird:
This PR has been rebased upon recent PRs, most importantly https://github.com/oneapi-src/oneAPI-spec/pull/506.

source/elements/oneMKL/source/domains/dft/descriptor.rst
hjabird: I've reworded it in [073c506](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/073c506d3f9dc468d6f464adae4278ee8f8ff2ad) - do you think this is clearer?

source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
```diff
@@ -0,0 +1,86 @@
...
+Workspace placement
+--------------------------------------
+
+DFT implementations often require temporary storage for intermediate data whilst computing DFTs.
+This temporary storage is referred to as a *workspace*.
+Whilst this is managed automatically by default (``config_param::WORKSPACE_AUTOMATIC``), 
+it may be preferable to provide an external workspace (``config_param::WORKSPACE_EXTERNAL``) for the following reasons:
+
+* To reduce the number of GPU mallocs / frees
+* To reduce memory consumption
```
Rbiessy, line 16: Following the discussion in https://github.com/oneapi-src/oneAPI-spec/pull/509#discussion_r1381333385 we could mention here that using `WORKSPACE_EXTERNAL` may hinder performance for some backends and configurations.

source/elements/oneMKL/source/domains/dft/descriptor.rst
Rbiessy: Yes, I have a preference for that version, thanks.

source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
hjabird: Done in [c75415e](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/c75415e9b9d3edeade96de944e76ead7e94d3479).

source/elements/oneMKL/source/domains/dft/descriptor.rst
```diff
@@ -417,6 +422,83 @@ type ``oneapi::mkl::dft::domain``, ``oneapi::mkl::dft::precision``,
...
+for compute calls until this function has been successfully called.
+
+The type of workspace must match the compute calls for which it is used.
+That is, if the workspace is provided as a ``sycl::buffer``, the compute
+calls must also use ``sycl::buffer`` for their arguments. Likewise, a USM
+allocated workspace must only be used with USM compute calls.
+Failing to do this will result in an invalid descriptor for compute calls.
+
+If the workspace is a USM allocation, the user must not use it for other purposes
+in parallel whilst the DFT `compute_forward` or `compute_backward` are in progress.
```
raphael-egan, line 450:
```suggestion
in parallel whilst the DFT ``compute_forward`` or ``compute_backward`` are in progress.
```
renders better in HTML (italic -> sans-serif type of police)

source/elements/oneMKL/source/domains/dft/descriptor.rst
```diff
@@ -417,6 +422,83 @@ type ``oneapi::mkl::dft::domain``, ``oneapi::mkl::dft::precision``,
...
+      A workspace buffer where ``scalar_type`` is the floating point type according to ``prec``. This buffer must be sufficiently large or an exception will be thrown.
+
+   workspaceUSM
+      A workspace USM allocation where ``scalar_type`` is the floating point type according to ``prec``. This allocation must be accessible on the device on which the descriptor is committed. It is assumed that this USM allocation is sufficiently large. The pointer is expected to be aligned to ``scalar_type``.
+
+.. container:: section
+
+   .. rubric:: Throws
+
+   The `descriptor::set_workspace()` routine shall throw the following exceptions if the associated condition is detected. An implementation may throw additional implementation-specific exception(s) in case of error conditions not covered here:
```
raphael-egan, line 491:
```suggestion
   The ``descriptor::set_workspace()`` routine shall throw the following exceptions if the associated condition is detected. An implementation may throw additional implementation-specific exception(s) in case of error conditions not covered here:
```
same thing: renders better in HTML + consistency with other similar sub-sections in this page.

source/elements/oneMKL/source/domains/dft/enums_and_config_params.rst
```diff
@@ -262,6 +265,16 @@ the :ref:`descriptor<onemkl_dft_descriptor>` class.
...
             | [0]
+      * -   :ref:`WORKSPACE_PLACEMENT<onemkl_dft_config_workspace_placement>`
+        -   Some FFT algorithm computation steps require a scratch space for permutations or other purposes. 
+            This parameter controls whether this scratch space is automatically allocated or provided by the user.
+        -   | :ref:`onemkl_dft_config_workspace_placement` (possible values are ``config_value::WORKSPACE_AUTOMATIC`` or ``config_value::WORKSPACE_EXTERNAL``).
+            | [``config_value::WORKSPACE_AUTOMATIC``]
+      * -   WORKSPACE_EXTERNAL_BYTES_REQUIRED
+        -   The required external workspace size in bytes when ``WORKSPACE_PLACEMENT`` is set to :ref:`config_value::WORKSPACE_EXTERNAL<onemkl_dft_config_value_workspace_automatic>`. 
+            A read-only value.
+        -   | ``std::int64_t``
```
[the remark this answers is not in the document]
hjabird: I think you're right in that it should be queryable - we state that a workspace can be set (but it may or may not do anything), even when `WORKSPACE_AUTOMATIC` is set. 

I don't think this should be used to query the internal memory requirements. For cases where external workspaces aren't supported (eg. MKL-CPU), we allow `WORKSPACE_EXTERNAL` to be set, but for the required space to always be zero. Even if we could obtain the workspace memory consumption of the descriptor in these cases, we wouldn't want to report it since the user might the allocate a non-zero external workspace that'd never be used.

source/elements/oneMKL/source/domains/dft/enums_and_config_params.rst
```diff
@@ -142,7 +142,10 @@ the :ref:`descriptor<onemkl_dft_descriptor>` class.
          OUTPUT_STRIDES, // deprecated
          
          FWD_DISTANCE,
-         BWD_DISTANCE
+         BWD_DISTANCE,
+
+         WORKSPACE_PLACEMENT,
+         WORKSPACE_EXTERNAL_BYTES_REQUIRED
```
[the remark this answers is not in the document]
hjabird: Following https://github.com/oneapi-src/oneAPI-spec/pull/509#discussion_r1383269938 I think it should still have `EXTERNAL` in the name. In that comment, I suggest a reason for the internal workspace allocation size not matching that reported by `WORKSPACE_EXTERNAL_BYTES_REQUIRED`. If we want to extend the interface for internal workspace size requirements later on (somehow!), which may be different from the value reported for this parameter, we should keep external in the name here to avoid confusion.

source/elements/oneMKL/source/domains/dft/enums_and_config_params.rst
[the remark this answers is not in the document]
hjabird: I've updated the doc string for `WORKSPACE_EXTERNAL_BYTES` in [e041739](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/e041739833d2e55d164382b70a0e48e6fe5530f1) to state that it is the minimum size allocation to use when calling `set_workspace`.

source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
```diff
@@ -0,0 +1,88 @@
...
+
+For some backends and configurations, ``config_param::WORKSPACE_EXTERNAL`` may reduce performance.
```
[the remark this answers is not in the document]
hjabird: Done in [749f614](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/749f6140d01036c076fa5c20a6600d74dd5ef931).

source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
```diff
@@ -0,0 +1,88 @@
...
+.. SPDX-License-Identifier: CC-BY-4.0
+
+.. _onemkl_dft_config_workspace_placement:
+
+Workspace placement
+--------------------------------------
+
+DFT implementations often require temporary storage for intermediate data whilst computing DFTs.
+This temporary storage is referred to as a *workspace*.
+Whilst this is managed automatically by default (``config_param::WORKSPACE_AUTOMATIC``),
```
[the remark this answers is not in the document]
hjabird: Done in [749f614](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/749f6140d01036c076fa5c20a6600d74dd5ef931).

source/elements/oneMKL/source/domains/dft/descriptor.rst
```diff
@@ -417,6 +422,83 @@ type ``oneapi::mkl::dft::domain``, ``oneapi::mkl::dft::precision``,
...
+
+.. container:: section
+
+   .. rubric:: Input Parameters
+
+   workspaceBuf
+      A workspace buffer where ``scalar_type`` is the floating point type according to ``prec``. This buffer must be sufficiently large or an exception will be thrown.
+
+   workspaceUSM
+      A workspace USM allocation where ``scalar_type`` is the floating point type according to ``prec``. This allocation must be accessible on the device on which the descriptor is committed. It is assumed that this USM allocation is sufficiently large. The pointer is expected to be aligned to ``scalar_type``.
```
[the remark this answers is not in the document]
hjabird: Done in [749f614](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/749f6140d01036c076fa5c20a6600d74dd5ef931).

source/elements/oneMKL/source/domains/dft/descriptor.rst
```diff
@@ -417,6 +422,83 @@ type ``oneapi::mkl::dft::domain``, ``oneapi::mkl::dft::precision``,
...
+      void descriptor<prec,dom>::set_workspace(scalar_type* workspaceUSM);
+
+   }
+
+.. container:: section
+
+   .. rubric:: Input Parameters
+
+   workspaceBuf
+      A workspace buffer where ``scalar_type`` is the floating point type according to ``prec``. This buffer must be sufficiently large or an exception will be thrown.
```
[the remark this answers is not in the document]
hjabird: Done in [749f614](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/749f6140d01036c076fa5c20a6600d74dd5ef931).

source/elements/oneMKL/source/domains/dft/descriptor.rst
```diff
@@ -417,6 +422,83 @@ type ``oneapi::mkl::dft::domain``, ``oneapi::mkl::dft::precision``,
...
+
+.. rubric:: Description
+
+This function sets the workspace to use when computing DFTs for when an
+external workspace is set. 
+This function may only be called after the descriptor has been committed.
+The size of the provided workspace must be equal to or larger than the required 
+workspace size obtained by calling ``descriptor<prec, dom>::get_value(config_param::WORKSPACE_EXTERNAL_BYTES_REQUIRED, &workspaceBytes)``.
+
+A descriptor where ``WORKSPACE_EXTERNAL`` is specified is not a valid descriptor
```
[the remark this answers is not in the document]
hjabird: Done in [749f614](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/749f6140d01036c076fa5c20a6600d74dd5ef931).

source/elements/oneMKL/source/domains/dft/descriptor.rst
```diff
@@ -417,6 +422,83 @@ type ``oneapi::mkl::dft::domain``, ``oneapi::mkl::dft::precision``,
...
+calls must also use ``sycl::buffer`` for their arguments. Likewise, a USM
+allocated workspace must only be used with USM compute calls.
+Failing to do this will result in an invalid descriptor for compute calls.
+
+If the workspace is a USM allocation, the user must not use it for other purposes
+in parallel whilst the DFT `compute_forward` or `compute_backward` are in progress.
+
+This function can be called on committed descriptors where the workspace placement
+is not ``config_value::WORKSPACE_EXTERNAL``. The provided workspace may or may not
+be used in compute calls. However, the aforementioned restrictions will still apply.
```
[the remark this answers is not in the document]
hjabird: Yes. Hopefully this is clearer in [e041739](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/e041739833d2e55d164382b70a0e48e6fe5530f1).

source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
```diff
@@ -0,0 +1,88 @@
...
+See :ref:`onemkl_dft_descriptor_set_workspace` and :ref:`onemkl_dft_typical_usage_of_workspace_external`.
+
+.. _onemkl_dft_typical_usage_of_workspace_external:
+
+Typical usage of ``WORKSPACE_EXTERNAL``
++++++++++++++++++++++++++++++++++++++++
+
+Usage of ``WORKSPACE_EXTERNAL`` typically involves the following order of operations:
+
+#. ``WORKSPACE_EXTERNAL`` is set for the uncommitted descriptor.
```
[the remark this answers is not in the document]
hjabird: Done in [525100d](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/525100dea2a01232c0e8d68b59fef63e5df5aabf).

source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
```diff
@@ -0,0 +1,88 @@
...
+WORKSPACE_PLACEMENT
++++++++++++++++++++
+
+For ``config_param::WORKSPACE_PLACEMENT``, valid configuration values are ``config_value::WORKSPACE_AUTOMATIC`` and ``config_value::WORKSPACE_EXTERNAL``.
+
+.. container:: section
+
+  .. _onemkl_dft_config_value_workspace_automatic:
+
+  .. rubric:: WORKSPACE_AUTOMATIC
```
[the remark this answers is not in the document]
hjabird: Yes for `_onemkl_dft_config_value_workspace_automatic` - otherwise I get *WARNING: Content block expected for the "container" directive; none found.* The indent of `..rubric:` doesn't give any difference in the page that I can see, nor does it affect warnings. 

I'm new to writing in restructured text (thank you for your comments!), so do you think I should be formatting this differently. Perhaps I should get rid of the `.. container:: section`s, in which case no indents are needed?

source/elements/oneMKL/source/domains/dft/descriptor.rst
hjabird: Done in [56bef76](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/56bef7645c0c64c503a94aabd0b62097dc7e098f).

source/elements/oneMKL/source/domains/dft/descriptor.rst
hjabird: Done in [56bef76](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/56bef7645c0c64c503a94aabd0b62097dc7e098f).

source/elements/oneMKL/source/domains/dft/descriptor.rst
```diff
@@ -417,6 +422,83 @@ type ``oneapi::mkl::dft::domain``, ``oneapi::mkl::dft::precision``,
...
+      template <oneapi::mkl::dft::precision prec, oneapi::mkl::dft::domain dom>
+      void descriptor<prec,dom>::set_workspace(scalar_type* workspaceUSM);
+
+   }
+
+.. container:: section
+
+   .. rubric:: Input Parameters
+
+   workspaceBuf
```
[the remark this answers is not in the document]
hjabird: I didn't consider this, and I don't understand how a 1D sub-buffer could cause problems. However, I've added the restriction that a sub-buffer cannot be used in [cb59477](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/cb594776fba3b73c325e9233abdc1ad6ddd30e39).

The reasoning is as follows:
* The use of sub-buffers is not common.
* Closed-source oneMKL is an important backend, but does not support this feature.
* The main library supporting `sycl::buffers` in the interface is the closed-source oneMKL. Therefore codes being ported to the oneMKL interface library, that depend on the buffer interface, are likely to have been using closed-source oneMKL. Therefore, ported codes probably already support this restriction.

source/elements/oneMKL/source/domains/dft/enums_and_config_params.rst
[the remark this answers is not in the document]
raphael-egan: Thank you!
Could you point me to the part about "we state that a workspace can be set (but it **may or may not** do anything), even when WORKSPACE_AUTOMATIC is set", please? Is that the documented behavior for some/many backends? I might have missed it but I find that a little odd.

If MKL-CPU is the only backend that does not support external workspace for now, wouldn't it be more suitable to throw an "unimplemented" exception if setting an external workspace with MKL-CPU and guarantee that the external workspace that was set is used, in other cases?

source/elements/oneMKL/source/domains/dft/descriptor.rst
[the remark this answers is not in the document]
raphael-egan: Thank you!
FYI, if you want to know/understand the limitation we face with that: the workspace that we're given is conceptually divided in non-overlapping chunks attributed to separate parts of the descriptors, e.g., twiddle tables, scratch space for transposition, scratch space of children descriptors, etc. When using buffers, it is done by creating sub-buffers off the given workspace. However, one cannot create a sub-buffer from another sub-buffer, hence the restriction.

source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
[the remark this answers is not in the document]
raphael-egan: I do see a slight difference in the sub-title indentation when removing the indent before `..rubric::` on my end, but it might be browser-dependent: I honestly don't know.
I would not claim any expertise in restructured text myself, my own efforts were heavily guided by trial-and-error attempts and very specific web searches. I must say that the purpose of the `.. container:: section` directives remains very unclear to me. I opted for removing them in my own changes when that did not affect the end result, personally.

source/elements/oneMKL/source/domains/dft/enums_and_config_params.rst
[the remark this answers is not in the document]
hjabird: I have *This function can be called on committed descriptors where the workspace placement is not config_value::WORKSPACE_EXTERNAL. The provided workspace may or may not be used in compute calls. However, the aforementioned restrictions will still apply.* in `set_workspace`. 

Regarding setting the workspace, when an external workspace was set pre-commit:
* In oneMKL and cuFFT, I believe setting the workspace causes the internally allocated workspace to be freed. 
* In rocFFT, the workspace is allocated and freed on each compute call by default. In oneMKL interface library, the workspace is set as an external workspace by the interface library to avoid this behaviour!
*  Many DFT libraries don't support external workspaces.
    * And this is what the *may not* bit of this is about.
    * I want it to be possible for the exact same code to work on all backends. 
        * The *may not* means that we can just ignore an external workspace when `set_workspace` is called on a backend that doesn't support it. This is in line with returning a zero-size required external workspace for these backends.

The rational for not stating *unsupported* is that:
* The user would require need a code-path to deal with external workspace not being supported.
* This extra codepath would presumably just not use an external workspace - we can do this internally anyway with very little effort.

source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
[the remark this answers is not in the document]
hjabird: I've removed the `..container` in [e139d35](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/e139d3564eb03f99e7d7d3eb5755e959b2a4414c), to follow the style of your previous PR. Hopefully this also removes the indentation.

source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
```diff
@@ -0,0 +1,88 @@
...
+When a descriptor is committed with ``config_value::WORKSPACE_EXTERNAL`` set, 
+the user must provide an external workspace. 
+See :ref:`onemkl_dft_descriptor_set_workspace` and :ref:`onemkl_dft_typical_usage_of_workspace_external`.
+
+.. _onemkl_dft_typical_usage_of_workspace_external:
+
+Typical usage of ``WORKSPACE_EXTERNAL``
++++++++++++++++++++++++++++++++++++++++
+
+Usage of ``WORKSPACE_EXTERNAL`` typically involves the following order of operations:
```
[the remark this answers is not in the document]
raphael-egan: Thanks for the stack-exchange reference: it's always helpful for non-native speakers (and appreciated) :-)
My motivation - biased from my own native language - was that list elements following a colon are not supposed to be capitalized and should end with a semi-colon. However, I understand that it is not necessarily transferrable to English as your link points out. That being said, I had suggested changes in another bullet list the same page, above: I let you judge what's best in your opinion in either case but my personal preference is to use the same convention for similar cases.

source/elements/oneMKL/source/domains/dft/enums_and_config_params.rst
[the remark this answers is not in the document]
raphael-egan: Thank you for the clarifications: that makes sense to me, personally!
Just requesting a confirmation that you meant "post-commit" and not "pre-commit" in that latest comment (thumbs-up if that was indeed a typo)?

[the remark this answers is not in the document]
hjabird: *Regarding setting the workspace, when an external workspace was set pre-commit:* is just wrong unfortunately. I should have said, *Regarding setting the workspace, when `WORKSPACE_EXTERNAL` was **not** set pre-commit (ie. `WORKSPACE_PLACEMENT` is `WORKSPACE_AUTOMATIC`):*.

Sorry for the confusion.

rscohn2:
Is this ready to merge?

zhiweij1 (commented):
LGTM

hjabird:
> Is this ready to merge?

Hello @rscohn2 , my understanding is that this should be presented to the Math SIG before it can be merged.

rscohn2:
@spencerpatty: Was this PR targeted for 1.3 release?

source/elements/oneMKL/source/domains/dft/config_params/workspace_placement.rst
```diff
@@ -0,0 +1,87 @@
...
+--------------------------------------
+
+DFT implementations often require temporary storage for intermediate data whilst computing DFTs.
+This temporary storage is referred to as a *workspace*.
+Whilst this is managed automatically by default (``config_value::WORKSPACE_AUTOMATIC`` 
+set for ``config_param::WORKSPACE_PLACEMENT``), it may be preferable to provide an external 
+workspace (``config_value::WORKSPACE_EXTERNAL`` set for ``config_param::WORKSPACE_PLACEMENT``) 
+for the following reasons:
+
+* to reduce the number of GPU mallocs / frees;
```
lhuot, line 17: I think we do not need to specify GPU here.

hjabird: Done in [84c52ba](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/84c52ba0e130300d8df2c4cb77b65008158209da).

source/elements/oneMKL/source/domains/dft/descriptor.rst
```diff
@@ -417,6 +422,84 @@ type ``oneapi::mkl::dft::domain``, ``oneapi::mkl::dft::precision``,
...
+
+   .. rubric:: Throws
+
+   The ``descriptor::set_workspace()`` routine shall throw the following exceptions if the associated condition is detected. An implementation may throw additional implementation-specific exception(s) in case of error conditions not covered here:
+   
+   :ref:`oneapi::mkl::invalid_argument()<onemkl_exception_invalid_argument>`
+      If the provided buffer ``workspaceBuf`` is not sufficiently large or is a sub-buffer, or if the provided USM allocation ``workspaceUSM`` is ``nullptr`` when an external workspace of size greater than zero is required.
+
+   :ref:`oneapi::mkl::uninitialized()<onemkl_exception_uninitialized>`
+      If ``set_workspace`` is called before the descriptor is committed.
```
Rbiessy, line 498: I can see that you're intentionally allowing users to call `set_workspace` even if `WORKSPACE_PLACEMENT` is set to `WORKSPACE_AUTOMATIC`. Shouldn't we throw a `oneapi::mkl::exception` in that case instead? My feeling is that the most likely scenario is that this would be a user mistake that intended to use external workspace and forgot to set `WORKSPACE_PLACEMENT`.

Rbiessy: That makes sense, thanks

source/elements/oneMKL/source/domains/dft/enums_and_config_params.rst
```diff
@@ -262,6 +265,16 @@ the :ref:`descriptor<onemkl_dft_descriptor>` class.
...
             | [0]
+      * -   :ref:`WORKSPACE_PLACEMENT<onemkl_dft_config_workspace_placement>`
+        -   Some FFT algorithm computation steps require a scratch space for permutations or other purposes. 
+            This parameter controls whether this scratch space is automatically allocated or provided by the user.
+        -   | :ref:`onemkl_dft_enum_config_value` (possible values are ``config_value::WORKSPACE_AUTOMATIC`` or ``config_value::WORKSPACE_EXTERNAL``).
+            | [``config_value::WORKSPACE_AUTOMATIC``]
+      * -   WORKSPACE_EXTERNAL_BYTES
+        -   The required minimum external workspace size for use by :ref:`set_workspace<onemkl_dft_descriptor_set_workspace>`.
+            A read-only value.
+        -   | ``std::int64_t``
```
hjabird, line 276: This should throw if called on an uncommitted descriptor.

source/elements/oneMKL/source/domains/dft/descriptor.rst
```diff
@@ -417,6 +422,84 @@ type ``oneapi::mkl::dft::domain``, ``oneapi::mkl::dft::precision``,
...
+
+A descriptor where ``config_value::WORKSPACE_EXTERNAL`` is specified for 
+``config_param::WORKSPACE_PLACEMENT`` is not a valid descriptor for compute 
+calls until this function has been successfully called.
+
+The type of workspace must match the compute calls for which it is used.
+That is, if the workspace is provided as a ``sycl::buffer``, the compute
+calls must also use ``sycl::buffer`` for their arguments. Likewise, a USM
+allocated workspace must only be used with USM compute calls.
+Failing to do this will result in an invalid descriptor for compute calls.
```
dbollapr, line 448: Was there any specific reason for this limitation?

source/elements/oneMKL/source/domains/dft/descriptor.rst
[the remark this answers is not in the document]
dbollapr: I believe we can lift this limitation in closed source oneMKL by using ranged accessors instead of sub-buffers. If so, do we still want to keep this limitation in the spec?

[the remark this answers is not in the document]
hjabird: I think we want to keep this limitation for now:
* This should work with the oneMKL that can be downloaded today.
* Its easier to remove this constraint in the future instead of deciding it needs to be added back in.
* I've not seen sub-buffers used that much, so I'd be interested to know how many people this limitation affects.

source/elements/oneMKL/source/domains/dft/descriptor.rst
hjabird: Yes. The oneMKL (MKL-GPU) backend has this limitation.

A second reason is that the oneMKL interface treats the `sycl::buffer` and USM somewhat orthogonally. The buffer interface does not accept or return `sycl::event`s for instance. Mixing USM and buffers with it would be an odd choice to my mind.

dbollapr: Right oneMKL does not support this. But I was trying to check that technically there is no limitation. I believe initially we wanted to support mixing in closed source but did not due to some eSIMD issues.

source/elements/oneMKL/source/domains/dft/descriptor.rst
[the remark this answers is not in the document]
dbollapr: Sounds good to me as we want it to work with oneMKL as of today.

source/elements/oneMKL/source/domains/dft/descriptor.rst
hjabird: Are you happy for this comment to be resolved?

hjabird:
Following the Math SIG, I tried to find examples uses of cufft workspace estimates on Github. Based on a small number of samples:
* `cufftestimate*` when used, often appears to have been used incorrectly. Users assume treat the value obtained as a conservative estimate, when it may return a smaller than the actual required work area.
* In many (most?) cases, the workspace requirement is obtained during plan creation using the `cufftMakePlan*` set of functions.

I don't think that pre-commit workspace size estimates are needed in this spec PR.

source/elements/oneMKL/source/domains/dft/enums_and_config_params.rst
hjabird: Addressed in [8d71156](https://github.com/oneapi-src/oneAPI-spec/pull/509/commits/8d71156cdb56d1645593168c20de1137e4fd18e5).

hjabird:
@lhuot @Rbiessy I believe this is now ready to merge. Having not been available for the Math SIG meeting, is there anything you think I've missed?

Rbiessy (approved):
I think this is ready as well.

source/elements/oneMKL/source/domains/dft/descriptor.rst
dbollapr: Yes, go ahead. Thank you

hjabird:
@rscohn2 I believe this is ready to merge. It has approval from both the Codeplay and Intel FastFour team, and has been presented at the Math SIG. I don't have the capability to merge this myself.

PR #512, opened 2023-11-04
[oneMKL][SpBLAS] Add `nnz` arg to `oneapi::mkl::sparse::set_csr_data()` API

gajanan-choudhary:
## Changes
* Adds `intType nnz` argument to `oneapi::mkl::sparse::set_csr_data()` API.
* Changes `num_rows`, `num_cols` to `nrows`, `ncols`, to also avoid using `num_nz` instead of `nnz` that we are proposing.
* Updates CSR matrix description with `nnz` definition, usage.

## Need
* cuSPARSE contains an `nnz` argument in their [cusparseCreateCsr() API](https://docs.nvidia.com/cuda/cusparse/index.html#cusparsecreatecsr). Adding the cuSPARSE backend when implementing the oneMKL specification will require workarounds if the `nnz` argument is not added to the specification. In most cases, nnz is known (except when it is being found as part of the API such as sparse + sparse = sparse or sparse * sparse = sparse matrix operations, but note that such APIs are currently not part of the oneAPI/oneMKL specification.)
* We would like this change to be part of the oneMKL spec version 1.3 that's to be released soon.


source/elements/oneMKL/source/domains/spblas/setcsrstructure.rst
```diff
@@ -128,8 +135,9 @@ set_csr_data (USM version)
 
       sycl::event set_csr_data (sycl::queue                           &queue,
                                 oneapi::mkl::sparse::matrix_handle_t  handle,
-                                intType                               num_rows,
-                                intType                               num_cols,
+                                intType                               nrows,
+                                intType                               ncols,
+                                intType                               nnz,
```
spencerpatty, line 140:
```suggestion
                                const intType                         nrows,
                                const intType                         ncols,
                                const intType                         nnz,
```

source/elements/oneMKL/source/domains/spblas/setcsrstructure.rst
```diff
@@ -33,8 +33,9 @@ set_csr_data (Buffer version)
 
       void set_csr_data (sycl::queue                           &queue,
                          oneapi::mkl::sparse::matrix_handle_t  handle,
-                         intType                               num_rows,
-                         intType                               num_cols,
+                         intType                               nrows,
+                         intType                               ncols,
+                         intType                               nnz,
```
spencerpatty, line 38:
```suggestion
                         const intType                         nrows,
                         const intType                         ncols,
                         const intType                         nnz,
```

spencerpatty: It seems like we should have const on these.  Is there an argument for not doing it ?

spencerpatty (approved):
pending decision about const or no on sizes, I approve.

source/elements/oneMKL/source/domains/spblas/setcsrstructure.rst
gajanan-choudhary: Intel oneMKL has `const`, but rocSPARSE, or cuSPARSE do not have a `const` in their CSR matrix creation APIs. See:
* [oneapi::mkl::sparse::set_csr_data()](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2023-2/oneapi-mkl-sparse-set-csr-data.html)
* [rocsparse_create_csr_descr()](https://rocm.docs.amd.com/projects/rocSPARSE/en/latest/auxiliary.html#rocsparse-create-csr-descr)
* [cusparseCreateCsr()](https://docs.nvidia.com/cuda/cusparse/index.html#cusparsecreatecsr)

Coming from C++, I think we should have a `const` in the spec, and so should, I believe, cuSPARSE and rocSPARSE. Because these are just integers, I don't think it matters either for functional correctness or performance reasons whether or not we have a `const` in there.

cuSPARSE/rocSPARSE backends only have `int64_t` type on nrows/ncols/nnz and not both {`int32_t`, `int64_t`} options, so libraries implementing this oneMKL Spec must anyway create a copy from `int32_t` to `int64_t` when dispatching to those backends, We might as well go ahead and add the `const` as you suggested here.

gajanan-choudhary: Actually on second thought, we also have the enum, `oneapi::mkl::index_base` in this API and additional enums in many other APIs which are not `const` (yet). We should be more disciplined and uniform about how and where we add the `const` keyword. I think in the next version of the spec, we should fix the `const`-ness of variables in all the APIs instead of just some. If only we had more time, we could have done it right in the upcoming v1.3. I'd like to keep it non-const for now and deal with this later.

gajanan-choudhary:
@rscohn2, @mkrainiuk, could one of you please merge this PR in by EOD today so it can go into spec v1.3? @spencerpatty reached out to me and said he might be unavailable today to merge the PR in.

PR #520, opened 2024-02-28
Fix gemm_batch documentation

AidanBeltonS:
The gemm_batch doc specifies that there are multiple typenames for alpha/beta, a, b, and c arguments. This is not reflected in the interface itself. This PR updates the template interface to reflect the table at the top of the file. This also brings it inline with the gemm interface.

A secondary issue I ran into is  `const sycl::span<std::int64_t> &alpha` seems incorrect as I believe it should be of type `Ts` or `T` and not an int. I think this might have been a typo.


AidanBeltonS:
@Alcpz, @mmeterel, cc

mmeterel:
@AidanBeltonS Thanks for catching and fixing the issues. Looks good to me.

PR #525, opened 2024-03-28
[oneMKL][RNG] Fixed some bugs for oneMKL RNG device API 

andreyfe1:
It doesn't make sense to keep constructors with `std::initializer_list` for `mcg31m1` and `mcg59` generators since one `seed` value is enough. Due to size of the state for `mcg31m1` uint32_t is enough to store it. For `mcg59` uint64_t is enough. So, let's keep only `std::uint32_t seed` for `mcg31m1` and `std::uint64_t seed` for `mcg59`.


aelizaro (approved):
Looks good to me, thank you @andreyfe1 for catching this issue!

PR #532, opened 2024-04-23
[oneMKL] fix gemv_batch op

mmeterel:
This PR is fixing #531


andrewtbarker (approved):
Looks good from my perspective.

PR #541, opened 2024-05-22
Fixed roadmap dates and table formatting

### LINKED ISSUE #540: All dates for (upcoming) 1.4 release are from 2023
#538 adds the schedule for the 1.4 release, but all of the dates listed are from 2023. Is that just a typo and the dates are correct but just a year off?
Thanks for adding the 1.4 release schedule!

Also, note that the two rows in the "Version" table are combined: https://github.com/uxlfoundation/oneAPI-spec/blob/main/roadmap.rst

rscohn2:
Thanks to @sknepper for pointing out the issues.

Fixes #540


sknepper (approved):
Looks great, thanks! I verified by looking at the file in your branch https://github.com/rscohn2/oneAPI-spec/blob/dev/fix-roadmap/roadmap.rst that the tables are appearing as expected.

PR #542, opened 2024-05-27
[oneMKL][spblas] Restrict features not supported by any backends

Rbiessy:
Update sparse specification to avoid features that are not supported by Intel oneMKL, cuSPARSE or rocSPARSE.
This is used in https://github.com/oneapi-src/oneMKL/pull/500

- Do not allow to change the sparse handle types as no backend support it currently.
- Change the restrictions on matrix view to avoid configurations that are not supported by any backend currently.
- Simplify constructor of matrix_view as the diagonal view is not useful currently. This could cause confusion.
- Do not allow spmv conjtrans with symmetric or Hermitian matrices

### DIFF
--- a/source/elements/oneMKL/source/domains/spblas/data_types/set_coo_matrix_data.rst
+++ b/source/elements/oneMKL/source/domains/spblas/data_types/set_coo_matrix_data.rst
@@ -53,11 +53,11 @@ set_coo_matrix_data (Buffer version)
 
    dataType
       See :ref:`supported template types<onemkl_sparse_supported_types>`. Must
-      be the same type than what was used when creating the ``matrix_handle_t``.
+      be the same type as was used when creating the ``matrix_handle_t``.
 
    indexType
       See :ref:`supported template types<onemkl_sparse_supported_types>`. Must
-      be the same type than what was used when creating the ``matrix_handle_t``.
+      be the same type as was used when creating the ``matrix_handle_t``.
 
 .. container:: section
 
@@ -150,12 +150,12 @@ set_coo_matrix_data (USM version)
    .. rubric:: Template parameters
 
    dataType
-      See :ref:`supported template types<onemkl_sparse_supported_types>`. Can be
-      a different type than what was used when creating the ``matrix_handle_t``.
+      See :ref:`supported template types<onemkl_sparse_supported_types>`. Must
+      be the same type as was used when creating the ``matrix_handle_t``.
 
    indexType
-      See :ref:`supported template types<onemkl_sparse_supported_types>`. Can be
-      a different type than what was used when creating the ``matrix_handle_t``.
+      See :ref:`supported template types<onemkl_sparse_supported_types>`. Must
+      be the same type as was used when creating the ``matrix_handle_t``.
 
 .. container:: section
 
--- a/source/elements/oneMKL/source/domains/spblas/data_types/set_coo_matrix_data.rst
+++ b/source/elements/oneMKL/source/domains/spblas/data_types/set_coo_matrix_data.rst
@@ -52,12 +52,12 @@ set_coo_matrix_data (Buffer version)
    .. rubric:: Template parameters
 
    dataType
-      See :ref:`supported template types<onemkl_sparse_supported_types>`. Can be
-      a different type than what was used when creating the ``matrix_handle_t``.
+      See :ref:`supported template types<onemkl_sparse_supported_types>`. Must
+      be the same type than what was used when creating the ``matrix_handle_t``.
 
    indexType
-      See :ref:`supported template types<onemkl_sparse_supported_types>`. Can be
-      a different type than what was used when creating the ``matrix_handle_t``.
+      See :ref:`supported template types<onemkl_sparse_supported_types>`. Must
+      be the same type than what was used when creating the ``matrix_handle_t``.
 
 .. container:: section
 
--- a/source/elements/oneMKL/source/domains/spblas/operations/spmm.rst
+++ b/source/elements/oneMKL/source/domains/spblas/operations/spmm.rst
@@ -273,6 +273,14 @@ spmm
      before ``spmm`` with the same arguments. ``spmm`` can then be called
      multiple times. Calling ``spmm_optimize`` on the same descriptor can reset
      some of the descriptor's data such as the ``workspace``.
+   - In the general case, not calling the functions in the order specified above
+     is undefined behavior. Not calling ``spmm_buffer_size`` or
+     ``spmm_optimize`` at least once will throw an
+     :ref:`oneapi::mkl::uninitialized<onemkl_exception_uninitialized>`
+     exception. Calling ``spmm`` with arguments not matching ``spmm_optimize``
+     will throw an
+     :ref:`oneapi::mkl::invalid_argument<onemkl_exception_invalid_argument>`
+     exception, unless stated otherwise.
    - The data of the dense handles ``B_handle`` and ``C_handle`` and the scalars
      ``alpha`` and ``beta`` can be reset before each call to ``spmm``. Changing
      the data of the sparse handle ``A_handle`` is undefined behavior.
--- a/source/elements/oneMKL/source/domains/spblas/operations/spmm.rst
+++ b/source/elements/oneMKL/source/domains/spblas/operations/spmm.rst
@@ -302,7 +302,9 @@ spmm
 
    A_view
       Specifies which part of the handle should be read as described by
-      :ref:`onemkl_sparse_matrix_view`.
+      :ref:`onemkl_sparse_matrix_view`. The ``type_view`` field must be
+      ``matrix_descr::general`` and the ``uplo_view`` and ``diag_view`` fields
+      are ignored.
 
    A_handle
       Sparse matrix handle object representing :math:`A`.
--- a/source/elements/oneMKL/source/domains/spblas/operations/spmm.rst
+++ b/source/elements/oneMKL/source/domains/spblas/operations/spmm.rst
@@ -275,7 +275,7 @@ spmm
      some of the descriptor's data such as the ``workspace``.
    - In the general case, not calling the functions in the order specified above
      is undefined behavior. Not calling ``spmm_buffer_size`` or
-     ``spmm_optimize`` at least once will throw an
+     ``spmm_optimize`` at least once with a given descriptor will throw an
      :ref:`oneapi::mkl::uninitialized<onemkl_exception_uninitialized>`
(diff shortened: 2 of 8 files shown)


source/elements/oneMKL/source/domains/spblas/data_types/set_coo_matrix_data.rst
```diff
@@ -52,12 +52,12 @@ set_coo_matrix_data (Buffer version)
    .. rubric:: Template parameters
 
    dataType
-      See :ref:`supported template types<onemkl_sparse_supported_types>`. Can be
-      a different type than what was used when creating the ``matrix_handle_t``.
+      See :ref:`supported template types<onemkl_sparse_supported_types>`. Must
+      be the same type than what was used when creating the ``matrix_handle_t``.
```
hjabird, line 56:
```suggestion
      See :ref:`supported template types<onemkl_sparse_supported_types>`. Must
      be the same type as was used when creating the ``matrix_handle_t``.
```

And similar elsewhere.

source/elements/oneMKL/source/domains/spblas/operations/spmm.rst
```diff
@@ -302,7 +302,9 @@ spmm
 
    A_view
       Specifies which part of the handle should be read as described by
-      :ref:`onemkl_sparse_matrix_view`.
+      :ref:`onemkl_sparse_matrix_view`. The ``type_view`` field must be
+      ``matrix_descr::general`` and the ``uplo_view`` and ``diag_view`` fields
+      are ignored.
```
hjabird, line 307: Is it useful to generate an error instead of ignoring?

source/elements/oneMKL/source/domains/spblas/data_types/set_coo_matrix_data.rst
Rbiessy: Thank you, I missed many references that I meant to change as well!
This is fixed in https://github.com/uxlfoundation/oneAPI-spec/pull/542/commits/1fdd42d6cb0c400b1ff7c4821a63d6bff1c79d4b

source/elements/oneMKL/source/domains/spblas/operations/spmm.rst
Rbiessy: We can't generate an error with the current API. The `uplo_view` and `diag_view` don't have anything like a `none` value. We would need to use `std::optional` in the matrix view but I think it would make the type a bit clunky.

Rbiessy:
Ping @gajanan-choudhary and @spencerpatty

source/elements/oneMKL/source/domains/spblas/operations/spsv.rst
```diff
@@ -283,8 +285,8 @@ spsv
 
    A_view
       Specifies which part of the handle should be read as described by
-      :ref:`onemkl_sparse_matrix_view`. ``A_view.type_view`` must be
-      ``matrix_descr::triangular`` or ``matrix_descr::diagonal``.
+      :ref:`onemkl_sparse_matrix_view`. The ``type_view`` field must be
+      ``matrix_descr::triangular``.
```
gajanan-choudhary, line 289: TRSV also does take in and support the `diagonal` field, so please add that in. Both oneMKL (see [oneapi::mkl::sparse::trsv()](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2024-1/oneapi-mkl-sparse-trsv.html)) and cuSPARSE (see [cusparseSpMatSetAttribute()](https://docs.nvidia.com/cuda/cusparse/#cusparsespmatsetattribute) that must be called before cusparseSpSV()) support that.

source/elements/oneMKL/source/domains/spblas/operations/spmm.rst
```diff
@@ -302,7 +304,9 @@ spmm
...
```
gajanan-choudhary, line 309:
```suggestion
      :ref:`onemkl_sparse_matrix_view`. The ``type_view`` field must be
      ``matrix_descr::general`` and the ``uplo_view`` and ``diag_view`` fields
      are currently ignored, but support may be expanded in the future.
```

source/elements/oneMKL/source/domains/spblas/operations/spmv.rst
```diff
@@ -292,7 +294,11 @@ spmv
 
    A_view
       Specifies which part of the handle should be read as described by
-      :ref:`onemkl_sparse_matrix_view`.
+      :ref:`onemkl_sparse_matrix_view`. The ``type_view`` field cannot be
+      ``matrix_descr::diagonal``. The ``diag_view`` field can be ``diag::unit``
+      if and only if ``type_view`` is ``matrix_descr::triangular``. The
+      ``type_view`` field cannot be ``matrix_descr::symmetric`` nor
+      ``matrix_descr::hermitian`` if ``opA`` is ``transpose::conjtrans``.
```
gajanan-choudhary, line 301:
```suggestion
      ``matrix_descr::hermitian`` if ``opA`` is ``transpose::conjtrans``.
      Support may be expanded in the future.
```

source/elements/oneMKL/source/domains/spblas/operations/spsv.rst
gajanan-choudhary: Also add in `Support may be expanded in the future`

Rbiessy: The `diag_view` field is supported when the `matrix_view` is `triangular`. The `diag_view` field is tested and works.
This change means that there is no way for the user to indicate to any of the backends that the input matrix only has values in the `diagonal`.
I don't like adding that support may be expanded in the future, this is true for pretty much all the limitations in the spec. I think this is just a detail that hinder readability.

source/elements/oneMKL/source/domains/spblas/operations/spmm.rst
Rbiessy: I have answered this in the comment above https://github.com/uxlfoundation/oneAPI-spec/pull/542#discussion_r1723195090

source/elements/oneMKL/source/domains/spblas/operations/spmv.rst
Rbiessy: I have answered this in the comment above https://github.com/uxlfoundation/oneAPI-spec/pull/542#discussion_r1723195090

Rbiessy:
Reverted the change https://github.com/uxlfoundation/oneAPI-spec/pull/542/commits/fec841e3886f0572a75ce3a9f05064ac3642ef14 to answer https://github.com/oneapi-src/oneMKL/pull/500#discussion_r1729273486.

source/elements/oneMKL/source/domains/spblas/operations/spsv.rst
gajanan-choudhary: Ignore the support to be added in the future; that's fine.

> This change means that there is no way for the user to indicate to any of the backends that the input matrix only has values in the diagonal.

What you mentioned would be a user-guaranteed `matrix_property`. Since we are talking about `matrix_view`s here, I believe you meant, "there is no way for user to indicate to any of the backends to read and use only the diagonal values (among others) in the input sparse matrix". That would be correct, because no backends support `matrix_descr type_view = matrix_descr::diagonal` in the `matrix_view` structure right now for any spec APIs.

You mentioned in `matrix_view.rst`:
>    Each operation documents which combination of ``type_view``, ``uplo_view`` and ``diag_view`` are valid.

That's why I am asking for us to add that `matrix_descr type_view = matrix_descr::triangular` and `diag diag_view = diag::{unit, nonunit}` must **both** be specified with `spsv`. What we have here in this PR is only about the ``type_view`` field; info about `diag diag_view=?` field is missing here.

gajanan-choudhary: Maybe there's some confusion: `diag diag_view` is not specific to `matrix_descr::diagonal` alone. `diag diag_view` is also applicable to and frequently used with `matrix_descr::triangular`.

gajanan-choudhary: What we need here is:
```
The ``type_view`` field must be ``matrix_descr::triangular``. The ``diag_view`` must be specified to ``diag::nonunit`` or ``diag::unit``
```
On a related note, at the bottom of the [Matrix view page](https://uxlfoundation.github.io/oneAPI-spec/spec/elements/oneMKL/source/domains/spblas/matrix_view.html#onemkl-sparse-matrix-descriptor), we have:
```
If the matrix_desc is diagonal, diag_view is initialized to diag::unit.
```
The default for `diag_view` should always be `nonunit`. We want users to intentionally choose `unit` because that is a rarer usage as opposed to `nonunit`. Do you mind fixing that in this PR?

source/elements/oneMKL/source/domains/spblas/operations/spmm.rst
```diff
@@ -273,9 +273,19 @@ spmm
      before ``spmm`` with the same arguments. ``spmm`` can then be called
      multiple times. Calling ``spmm_optimize`` on the same descriptor can reset
      some of the descriptor's data such as the ``workspace``.
+   - In the general case, not calling the functions in the order specified above
+     is undefined behavior. Not calling ``spmm_buffer_size`` or
+     ``spmm_optimize`` at least once will throw an
```
gajanan-choudhary, line 278: Maybe the following, since that's how you have also implemented it in https://github.com/oneapi-src/oneMKL/pull/500?
```suggestion
     ``spmm_optimize`` at least once with a given descriptor will throw an
```

Rbiessy: Done in https://github.com/uxlfoundation/oneAPI-spec/pull/542/commits/b3eb26b577e60ea089e47cd306468c194ce8bc63

source/elements/oneMKL/source/domains/spblas/operations/spsv.rst
Rbiessy: My intention was to only document when a field is ignored or when it cannot be set to a value but it could maybe cause some confusion. I've added a bit of clarification in https://github.com/uxlfoundation/oneAPI-spec/pull/542/commits/96b30890d61c52b31a88bb616faa80f1e198193c.
It's not quite your suggestion of "diag_view must be specified" since the field is already defaulted to `nonunit` see [here](https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneMKL/source/domains/spblas/matrix_view.rst?plain=1#L68).

gajanan-choudhary (approved):
LGTM!

PR #546, opened 2024-06-17
[oneDPL] Histogram APIs

danhoeflinger:
Proposing `oneapi::dpl::histogram` specification for oneDPL.

### DIFF
--- a/source/elements/oneDPL/source/parallel_api.rst
+++ b/source/elements/oneDPL/source/parallel_api.rst
@@ -524,39 +524,29 @@ The elements of ``[start, end)`` must be partitioned with respect to the compara
         typename ValueType>
     OutputIt
     histogram(Policy&& exec, InputIt start, InputIt end, Size num_bins,
-        ValueType first_bin_min_val, ValueType last_bin_max_val, OutputIt histogram_first)
-
-:code:`oneapi::dpl::histogram` performs a histogram operation over the data in :code:`[start, end)`
-into :code:`num_bins` bins stored in a sequence which starts with :code:`histogram_first`. The
-elements of :code:`[start, end)` are mapped into evenly divided bins between :code:`first_bin_min_val`
-and :code:`last_bin_max_val` such that an input element :code:`*(start + i)` maps to a bin index
-:code:`j` such that
-:code:`j = floor((*(start + i) - first_bin_min_val) / ((last_bin_max_val - first_bin_min_val) / num_bins)))`.
-The histogram operation counts the number of elements mapped to each bin and stores the count into
-:code:`*(histogram_first + j)` within the sequence :code:`[histogram_first, histogram_first + num_bins)`.
-Input values which do not map to a defined bin are skipped silently. Sufficient output data to store each
-bin must be provided, and :code:`Size` must be sufficient to store the counts of the histogram without overflow.
-All input and output sequences must be ``RandomAccessIterators``.
-
-
-.. code:: cpp
+        ValueType first_bin_min_val, ValueType last_bin_max_val, OutputIt histogram_first) // (1)
 
     template <typename Policy, typename InputIt1, typename InputIt2, typename OutputIt>
     OutputIt
     histogram(Policy&& exec, InputIt1 start, InputIt1 end, InputIt2 boundary_start,
-              InputIt2 boundary_end, OutputIt histogram_first)
+              InputIt2 boundary_end, OutputIt histogram_first)                             // (2)
 
-:code:`oneapi::dpl::histogram` performs a histogram operation over the data in :code:`[start, end)`
-into :code:`num_bins` bins stored in a sequence which starts with :code:`histogram_first`. The
-elements of :code:`[start, end)` are  mapped into bins as defined by their boundaries in
-:code:`[boundary_start, boundary_end)` such that an input element :code:`*(start + i)` maps to a bin
-index :code:`j` if and only if
-:code:`(boundary_start[j] <= *(start + i)) && (*(start + i) < boundary_start[j + 1])`.
-The histogram operation counts the number of elements mapped to each bin and stores the count into
-:code:`*(histogram_first + j)` within the sequence :code:`[histogram_first, histogram_first + num_bins)`.
-Input values which do not map to a defined bin are skipped silently. Sufficient output data to store each
-bin must be provided, and :code:`Size` must be sufficient to store the counts of the histogram without overflow.
+:code:`oneapi::dpl::histogram` performs a histogram operation over the data in :code:`[start, end)`,
+which counts the number of elements which map to a each of a set of bins and stores the count into
+a sequence in :code:`[histogram_first, histogram_first + num_bins)`. Input values which do not map
+to a defined bin are skipped silently. Sufficient output data to store each bin must be provided,
+and :code:`Size` must be sufficient to store the counts of the histogram without overflow.
 All input and output sequences must be ``RandomAccessIterators``.
+
+(1) The elements of :code:`[start, end)` are mapped into evenly divided bins between
+:code:`first_bin_min_val` and :code:`last_bin_max_val` such that an input element :code:`start[i]`
+maps into a bin :code:`histogram_first[j]` where
+:code:`j = floor((start[i] - first_bin_min_val) / ((last_bin_max_val - first_bin_min_val) / num_bins)))`.
+
+(2) The elements of :code:`[start, end)` are mapped into bins as defined by bin boundaries in
+:code:`[boundary_start, boundary_end)` such that an input element :code:`start[i]` maps to a bin
+:code:`histogram_first[j]` if and only if
+:code:`(boundary_start[j] <= start[i]) && (start[i] < boundary_start[j + 1])`.
 
 .. _`C++ Standard`: https://isocpp.org/std/the-standard
 .. _`SYCL`: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html
--- a/source/elements/oneDPL/source/parallel_api.rst
+++ b/source/elements/oneDPL/source/parallel_api.rst
@@ -537,7 +537,6 @@ by counting the number of elements that map to each of a set of bins and storing
 :code:`[histogram_first, histogram_first + num_bins)`. Input values that do not map
 to a defined bin are skipped silently. Sufficient output data to store each count must be provided,
 and :code:`Size` must be sufficient to store the counts of the histogram without overflow.
-All input and output sequences must be ``RandomAccessIterators``.
 
 1. The elements of :code:`[start, end)` are mapped into evenly divided bins between
 :code:`first_bin_min_val` and :code:`last_bin_max_val` such that an input element :code:`start[i]`
--- a/source/elements/oneDPL/source/parallel_api.rst
+++ b/source/elements/oneDPL/source/parallel_api.rst
@@ -536,7 +536,8 @@ The elements of ``[start, end)`` must be partitioned with respect to the compara
 by counting the number of elements that map to each of a set of bins and storing the counts into
 the output sequence starting from :code:`histogram_first`. Input values that do not map to a defined
 bin are skipped silently. The value type of :code:`OutputIt` must be an integral type of sufficient
-size to store the counts of the histogram without overflow.
+size to store the counts of the histogram without overflow. The return value is an iterator targeting
+past the last element of the output sequence starting from :code:`histogram_first`.
 
 1. The elements of :code:`[start, end)` are mapped into :code:`num_intervals` bins that evenly divide the range
    :code:`[first_interval_begin, last_interval_end)`. Each bin is of size
--- a/source/elements/oneDPL/source/parallel_api.rst
+++ b/source/elements/oneDPL/source/parallel_api.rst
@@ -524,9 +524,9 @@ The elements of ``[start, end)`` must be partitioned with respect to the compara
     template <typename Policy, typename InputIt, typename Size, typename OutputIt>
     OutputIt
     histogram(Policy&& exec, InputIt start, InputIt end, Size num_intervals,
-        typename ::std::iterator_traits<InputIt>::value_type first_interval_begin,
-        typename ::std::iterator_traits<InputIt>::value_type last_interval_end,
-        OutputIt histogram_first); // (1)
+              typename std::iterator_traits<InputIt>::value_type first_interval_begin,
+              typename std::iterator_traits<InputIt>::value_type last_interval_end,
+              OutputIt histogram_first); // (1)
 
     template <typename Policy, typename InputIt1, typename InputIt2, typename OutputIt>
     OutputIt
--- a/source/elements/oneDPL/source/parallel_api.rst
+++ b/source/elements/oneDPL/source/parallel_api.rst
@@ -524,7 +524,7 @@ The elements of ``[start, end)`` must be partitioned with respect to the compara
     template <typename Policy, typename InputIt, typename Size, typename OutputIt,
         typename ValueType>
     OutputIt
-    histogram(Policy&& exec, InputIt start, InputIt end, Size num_bins,
+    histogram(Policy&& exec, InputIt start, InputIt end, Size num_intervals,
         ValueType first_interval_begin, ValueType last_interval_end, OutputIt histogram_first); // (1)
 
     template <typename Policy, typename InputIt1, typename InputIt2, typename OutputIt>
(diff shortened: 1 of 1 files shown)


source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,35 @@ than an element in the range being searched.
...
+which counts the number of elements that map to each of a set of bins and stores the count into
+a sequence in :code:`[histogram_first, histogram_first + num_bins)`. Input values that do not map
+to a defined bin are skipped silently. Sufficient output data to store each count must be provided,
+and :code:`Size` must be sufficient to store the counts of the histogram without overflow.
+All input and output sequences must be ``RandomAccessIterators``.
+
+1. The elements of :code:`[start, end)` are mapped into evenly divided bins between
+:code:`first_bin_min_val` and :code:`last_bin_max_val` such that an input element :code:`start[i]`
+maps into a bin :code:`histogram_first[j]` where
+:code:`j = floor((start[i] - first_bin_min_val) / ((last_bin_max_val - first_bin_min_val) / num_bins)))`.
```
akukanov, line 714: First, this is quite hard to decipher.
Second, does it describe the math or the code to be used by an implementation? At least as it comes to division, there is difference between the two.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,40 @@ than an element in the range being searched.
...
+              InputIt2 boundary_end, OutputIt histogram_first);                             // (2)
+
+:code:`oneapi::dpl::histogram` computes the histogram over the data in :code:`[start, end)`
+by counting the number of elements that map to each of a set of bins and storing the counts into
+:code:`[histogram_first, histogram_first + num_bins)`. Input values that do not map to a defined
+bin are skipped silently. Sufficient output data to store each count must be provided, and the
+value type of :code:`OutputIt` must be an integral type sufficient to store the counts of the
+histogram without overflow.
+
+1. The elements of :code:`[start, end)` are mapped into bins which are evenly divided between
```
akukanov, line 711: > bins which are evenly divided between...

Should it be `bins that evenly divide (the range)` or maybe `bins which are evenly distributed between...`? Or am I nitpicking? :)

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,40 @@ than an element in the range being searched.
 
 The elements e of [start, end) must be partitioned with respect to the comparator used.
 
+.. code:: cpp
+
+    template <typename Policy, typename InputIt, typename Size, typename OutputIt,
+        typename ValueType>
+    OutputIt
+    histogram(Policy&& exec, InputIt start, InputIt end, Size num_bins,
+        ValueType first_bin_min_val, ValueType last_bin_max_val, OutputIt histogram_first); // (1)
```
akukanov, line 697: A potential ambiguity: will values equal to `last_bin_max_val` be counted? The parameter name kind of suggests these will. but the math below tells the opposite.

I would probably not use the term "bin" at all, changing it everywhere to "interval". The parameters could then be named `first_interval_begin` and `last_interval_end`. Perhaps still ambiguous, because mathematically these are half-intervals; yet it sounds better to me.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,40 @@ than an element in the range being searched.
...
+
+    template <typename Policy, typename InputIt1, typename InputIt2, typename OutputIt>
+    OutputIt
+    histogram(Policy&& exec, InputIt1 start, InputIt1 end, InputIt2 boundary_start,
+              InputIt2 boundary_end, OutputIt histogram_first);                             // (2)
+
+:code:`oneapi::dpl::histogram` computes the histogram over the data in :code:`[start, end)`
+by counting the number of elements that map to each of a set of bins and storing the counts into
+:code:`[histogram_first, histogram_first + num_bins)`. Input values that do not map to a defined
+bin are skipped silently. Sufficient output data to store each count must be provided, and the
```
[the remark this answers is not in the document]
akukanov: It's "not that simple" for several other algorithms too, e.g. for `reduce_by_segment`.

> For the (1), it is `num_bins`, but for (2) it is one `(boundary_end - boundary_start)`.

You can "define" `num_bins` being equal to `(boundary_end - boundary_start)` for the overload (2). As an example, here is a quote from the C++ standard:

> Let `last2` be `first2 + (last1 - first1)` for the overloads with no parameter `last2`.

source/elements/oneDPL/source/parallel_api.rst
danhoeflinger: changed.

source/elements/oneDPL/source/parallel_api.rst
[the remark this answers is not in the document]
danhoeflinger: I've added a brief listing of the number of bins to each algorithm, and removed this line in the intro.

source/elements/oneDPL/source/parallel_api.rst
danhoeflinger: No problem in nitpicking.  I agree that the language you propose here is better.  I've updated it.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,40 @@ than an element in the range being searched.
...
+value type of :code:`OutputIt` must be an integral type sufficient to store the counts of the
+histogram without overflow.
+
+1. The elements of :code:`[start, end)` are mapped into bins which are evenly divided between
+:code:`first_bin_min_val` and :code:`last_bin_max_val`. Each bin is of size
```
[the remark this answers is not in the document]
danhoeflinger: This should be fixed now.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,39 @@ than an element in the range being searched.
...
+1. The elements of :code:`[start, end)` are mapped into :code:`num_bins` bins that evenly divide the range
+   :code:`[first_interval_begin, last_interval_end)`. Each bin is of size
+   :code:`bin_size = (last_interval_end - first_interval_begin) / num_bins` as represented by a real
+   number without rounding or truncation. An input element maps to a bin :code:`histogram_first[j]`
+   if and only if
+   :code:`(first_interval_begin + j * bin_size <= start[i]) && (start[i] < first_interval_begin + (j + 1) * bin_size)`.
+   `ValueType` must be an arithmetic type. The value type of :code:`InputIt` must be an arithmetic type.
+
+2. The elements of :code:`[start, end)` are mapped into :code:`(boundary_end - boundary_start - 1)`
+   bins as defined by bin boundaries in :code:`[boundary_start, boundary_end)` such that an input
```
akukanov, line 719:
```suggestion
2. The elements of :code:`[start, end)` are mapped into :code:`(boundary_end - boundary_start - 1)`
   bins defined by the values in :code:`[boundary_start, boundary_end)`. An input 
```

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,39 @@ than an element in the range being searched.
...
+        ValueType first_interval_begin, ValueType last_interval_end, OutputIt histogram_first); // (1)
+
+    template <typename Policy, typename InputIt1, typename InputIt2, typename OutputIt>
+    OutputIt
+    histogram(Policy&& exec, InputIt1 start, InputIt1 end, InputIt2 boundary_start,
+              InputIt2 boundary_end, OutputIt histogram_first);                             // (2)
+
+:code:`oneapi::dpl::histogram` computes the histogram over the data in :code:`[start, end)`
+by counting the number of elements that map to each of a set of bins and storing the counts into
+:code:`[histogram_first, histogram_first + num_bins)`. Input values that do not map to a defined
```
akukanov, line 706:
```suggestion
by counting the number of elements that map to each of a set of bins and storing the counts into
the output sequence starting from :code:`histogram_first`. Input values that do not map to a defined
```

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,39 @@ than an element in the range being searched.
...
+   number without rounding or truncation. An input element maps to a bin :code:`histogram_first[j]`
+   if and only if
+   :code:`(first_interval_begin + j * bin_size <= start[i]) && (start[i] < first_interval_begin + (j + 1) * bin_size)`.
+   `ValueType` must be an arithmetic type. The value type of :code:`InputIt` must be an arithmetic type.
+
+2. The elements of :code:`[start, end)` are mapped into :code:`(boundary_end - boundary_start - 1)`
+   bins as defined by bin boundaries in :code:`[boundary_start, boundary_end)` such that an input 
+   element :code:`start[i]` maps to a bin :code:`histogram_first[j]` if and only if
+   :code:`(boundary_start[j] <= start[i]) && (start[i] < boundary_start[j + 1])`.  The value types
+   of :code:`InputIt1` and :code:`InputIt2` must be arithmetic types. The elements of
```
akukanov, line 722:
```suggestion
   of :code:`InputIt1` and :code:`InputIt2` must be arithmetic types. The values in
```
just to minimize confusion, as above "elements" refer to the input data.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,39 @@ than an element in the range being searched.
 
 The elements e of [start, end) must be partitioned with respect to the comparator used.
 
+.. code:: cpp
+
+    template <typename Policy, typename InputIt, typename Size, typename OutputIt,
+        typename ValueType>
+    OutputIt
+    histogram(Policy&& exec, InputIt start, InputIt end, Size num_bins,
```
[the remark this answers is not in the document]
danhoeflinger: done, thanks.

akukanov (commented):
Let's also add the description of the return value.

danhoeflinger:
> Let's also add the description of the return value.
added.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -705,7 +705,8 @@ The elements e of [start, end) must be partitioned with respect to the comparato
 by counting the number of elements that map to each of a set of bins and storing the counts into
 the output sequence starting from :code:`histogram_first`. Input values that do not map to a defined
 bin are skipped silently. The value type of :code:`OutputIt` must be an integral type of sufficient
-size to store the counts of the histogram without overflow.
+size to store the counts of the histogram without overflow. The return value is an iterator targeting
+past the last element of the output sequence starting from :code:`histogram_first`.
```
akukanov, line 709:
```suggestion
past the last stored count in the output sequence.
```

akukanov (approved):

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -519,5 +519,40 @@ than an element in the range being searched.
 
 The elements of ``[start, end)`` must be partitioned with respect to the comparator used.
 
+.. code:: cpp
+
+    template <typename Policy, typename InputIt, typename Size, typename OutputIt,
```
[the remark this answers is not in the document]
danhoeflinger: I believe you have caught something here that is a problem.  
I have a proposal to fix it, but I think we want to do more than just reorder the template parameters.

Instead of having a free floating `ValueType` template parameter.  I believe we should instead directly use `std::iterator_traits<InputIt>::value_type` as the type of `first_interval_begin` and `last_interval_end`.  

I think it is better for force users to provide bounds which are convertible to the `value_type` of the input sequence, and do the conversion once when accepting the arguments rather than each time in the comparison operator.  This will change the signature and some of the language.  I have updated it accordingly, please take a look.  

Also, while this is not the same exact API as what is currently implemented in oneDPL, it is not incompatible with the existing implementation's API.  I believe we can continue to support that API as an extension without a breaking change, but lets discuss that in the implementation PR, which will be up shortly.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -519,5 +519,41 @@ than an element in the range being searched.
...
 The elements of ``[start, end)`` must be partitioned with respect to the comparator used.
 
+.. code:: cpp
+
+    template <typename Policy, typename InputIt, typename Size, typename OutputIt>
+    OutputIt
+    histogram(Policy&& exec, InputIt start, InputIt end, Size num_intervals,
+        typename ::std::iterator_traits<InputIt>::value_type first_interval_begin,
+        typename ::std::iterator_traits<InputIt>::value_type last_interval_end,
+        OutputIt histogram_first); // (1)
```
akukanov, line 529:
```suggestion
    histogram(Policy&& exec, InputIt start, InputIt end, Size num_intervals,
              typename std::iterator_traits<InputIt>::value_type first_interval_begin,
              typename std::iterator_traits<InputIt>::value_type last_interval_end,
              OutputIt histogram_first); // (1)
```

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -519,5 +519,41 @@ than an element in the range being searched.
...
+    OutputIt
+    histogram(Policy&& exec, InputIt start, InputIt end, Size num_intervals,
+              typename std::iterator_traits<InputIt>::value_type first_interval_begin,
+              typename std::iterator_traits<InputIt>::value_type last_interval_end,
+              OutputIt histogram_first); // (1)
+
+    template <typename Policy, typename InputIt1, typename InputIt2, typename OutputIt>
+    OutputIt
+    histogram(Policy&& exec, InputIt1 start, InputIt1 end, InputIt2 boundary_start,
+              InputIt2 boundary_end, OutputIt histogram_first);                             // (2)
```
akukanov, line 534: Does it make sense to align `// (1)` and `// (2)` so that both are well visible?

danhoeflinger: yes, I'll fix this.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -519,5 +519,41 @@ than an element in the range being searched.
...
+   number without rounding or truncation. An input element ``start[i]`` maps to a bin
+   ``histogram_first[j]`` if and only if
+   ``(first_interval_begin + j * bin_size <= start[i]) && (start[i] < first_interval_begin + (j + 1) * bin_size)``.
+   The value type of ``InputIt`` must be an arithmetic type.
+
+2. The elements of ``[start, end)`` are mapped into ``(boundary_end - boundary_start - 1)``
+   bins defined by the values in ``[boundary_start, boundary_end)``. An input 
+   element ``start[i]`` maps to a bin ``histogram_first[j]`` if and only if
+   ``(boundary_start[j] <= start[i]) && (start[i] < boundary_start[j + 1])``.  The value types
+   of ``InputIt1`` and ``InputIt2`` must be arithmetic types. The values in
```
danhoeflinger, line 555: With the change to (1), do we want to require that `InputIt1` and `InputIt2` have matching value types for consistency between overloads?

I think the answer is probably no and that it is unnecessary, but I want to raise the question.  I also don't think this is incompatible with oneDPL's implementation, and oneDPL could just support more than the spec requires here without changing its interface or adding any static assertions.

akukanov: The real requirement seems to be that the types are "totally ordered with" (see https://en.cppreference.com/w/cpp/concepts/totally_ordered). I am not sure how to express it prior to C++20 concepts, though.

danhoeflinger: I believe we have "totally ordered with" from the fact that both value types are arithmetic types.  
What I'm asking is:  
Since (1) now dictates in the spec that the boundary min and max is the input data's value type, do we want to do something similar in (2), requiring that the boundaries specified match the input data's value type? 
I think it is a bit awkward to do this via some static assert or meta programming which is less clear from the signature itself.  Therefore, I think its enough that they are just both arithmetic types (which I believe makes them totally ordered with each other).

akukanov: I agree - I do not see why we would require the same type, given that the algorithm does not de facto need this.

danhoeflinger: OK calling it resolved.  Thanks.

akukanov (approved):

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -519,5 +519,41 @@ than an element in the range being searched.
...
+
+1. The elements of ``[start, end)`` are mapped into ``num_intervals`` bins that evenly divide the range
+   ``[first_interval_begin, last_interval_end)``. Each bin is of size
+   ``bin_size = (last_interval_end - first_interval_begin) / num_intervals`` as represented by a real
+   number without rounding or truncation. An input element ``start[i]`` maps to a bin
+   ``histogram_first[j]`` if and only if
+   ``(first_interval_begin + j * bin_size <= start[i]) && (start[i] < first_interval_begin + (j + 1) * bin_size)``.
+   The value type of ``InputIt`` must be an arithmetic type.
+
+2. The elements of ``[start, end)`` are mapped into ``(boundary_end - boundary_start - 1)``
```
akukanov, line 551:
```suggestion
2. The elements of ``[start, end)`` are mapped into ``std::distance(boundary_start, boundary_end) - 1)``
```

danhoeflinger:
We have decided to instead go with #571  in an offline discussion.  closing this.

PR #550, opened 2024-06-26
[oneDPL] Improve requirements for iterator source types

danhoeflinger:
This PR aims to improve the specification of `permutation_iterator`, `zip_iterator`, and `transform_iterator` by adding requirements for their source iterator types.  

The source iterator type for each of these iterators share the same requirements, for which a name requirement `ValidParallelIteratorSource` has been defined.  There may be a better or more concise name for this requirement.

`permutation_iterator` has a type `IndexMap` which has a modified set of requirements from `ValidParallelIteratorSource`, because some types like `zip_iterator` and `discard_iterator` do not make sense in this setting. Additionally, `IndexMap` can instead be a functor which transforms consecutive indices into a map.  In practice, one implementation of this functionality would be to use a `transform_iterator<counting_iterator, IndexMap>` to provide the iterator serving as the `IndexMap` in a similar way to other incoming `IndexMap` iterators.

### DIFF
--- a/source/elements/oneDPL/source/parallel_api.rst
+++ b/source/elements/oneDPL/source/parallel_api.rst
@@ -371,7 +371,7 @@ using the source iterator and unary function object provided.
 
 ``zip_iterator`` is an iterator-like type defined over one or more iterators. When dereferenced,
 the value returned from ``zip_iterator`` is a tuple of the values returned by dereferencing the
-source iterators over which the ``zip_iterator`` is defined.  The arithmetic operators of
+source iterators over which the ``zip_iterator`` is defined. The arithmetic operators of
 ``zip_iterator`` update the source iterators of a ``zip_iterator`` instance as though the
 operation were applied to each of these iterators. The types ``T`` within the template pack 
 ``Iterators...`` must satisfy ``AdaptingIteratorSource``.
--- a/source/elements/oneDPL/source/parallel_api.rst
+++ b/source/elements/oneDPL/source/parallel_api.rst
@@ -96,13 +96,13 @@ Let us define a named requirement, :code:`AdaptingIteratorSource`, to describe v
 types that can be used as source for oneDPL iterators as described below.
 The type :code:`Iter` satisfies the :code:`AdaptingIteratorSource` named requirement if it is any of the following:
 
- * a random access iterator
- * the unspecified iterator-like type returned by :code:`oneapi::dpl::begin` or :code:`oneapi::dpl::end`
- * a :code:`permutation_iterator`
- * a :code:`transform_iterator`
- * a :code:`counting_iterator`
- * a :code:`discard_iterator`
- * a :code:`zip_iterator`
+ * A random access iterator
+ * The unspecified iterator-like type returned by :code:`oneapi::dpl::begin` or :code:`oneapi::dpl::end`
+ * A :code:`permutation_iterator`
+ * A :code:`transform_iterator`
+ * A :code:`counting_iterator`
+ * A :code:`discard_iterator`
+ * A :code:`zip_iterator`
 
 .. code:: cpp
 
@@ -238,12 +238,12 @@ iterator instances to determine their position in the index map. :code:`SourceIt
 
 The type :code:`IndexMap` must be one of the following:
 
- * a random access iterator
- * the unspecified iterator-like type returned by :code:`oneapi::dpl::begin` or :code:`oneapi::dpl::end`
- * a :code:`permutation_iterator`
- * a :code:`transform_iterator`
- * a :code:`counting_iterator`
- * a functor with a signature equivalent to :code:`T operator()(const T&) const` where :code:`T` is a
+ * A random access iterator
+ * The unspecified iterator-like type returned by :code:`oneapi::dpl::begin` or :code:`oneapi::dpl::end`
+ * A :code:`permutation_iterator`
+ * A :code:`transform_iterator`
+ * A :code:`counting_iterator`
+ * A functor with a signature equivalent to :code:`T operator()(const T&) const` where :code:`T` is a
    :code:`std::iterator_traits<SourceIterator>::difference_type`
 
 
--- a/source/elements/oneDPL/source/parallel_api.rst
+++ b/source/elements/oneDPL/source/parallel_api.rst
@@ -96,7 +96,7 @@ Let us define a named requirement, :code:`ValidParallelIteratorSource`, to descr
 types that can be used as source for oneDPL's iterators as described below.
 The type :code:`Iter` satisfies the :code:`ValidParallelIteratorSource` named requirement if it satisfies
 at least one of the following:
- * :code:`Iter` satisfies the C++ named requirement :code:`LegacyRandomAccessIterator`
+ * :code:`Iter` is a random access iterator
  * :code:`Iter` is the unspecified iterator-like type returned by :code:`oneapi::dpl::begin` or :code:`oneapi::dpl::end`
  * :code:`Iter` is a valid :code:`permutation_iterator`
  * :code:`Iter` is a valid :code:`transform_iterator`
@@ -237,7 +237,7 @@ iterator instances to determine their position in the index map. :code:`SourceIt
 :code:`ValidParallelIteratorSource`.
 
 The type :code:`IndexMap` must satisfy at least one of the following:
- * :code:`IndexMap` satisfies the C++ named requirement :code:`LegacyRandomAccessIterator`
+ * :code:`IndexMap` is a random access iterator
  * :code:`IndexMap` is the unspecified iterator-like type returned by :code:`oneapi::dpl::begin` or :code:`oneapi::dpl::end`
  * :code:`IndexMap` is a valid :code:`permutation_iterator`
  * :code:`IndexMap` is a valid :code:`transform_iterator`
--- a/source/elements/oneDPL/source/parallel_api.rst
+++ b/source/elements/oneDPL/source/parallel_api.rst
@@ -92,17 +92,17 @@ Iterators
 The oneDPL iterators are defined in the ``<oneapi/dpl/iterator>`` header,
 in ``namespace oneapi::dpl``.
 
-Let us define a named requirement, :code:`AdaptingIteratorSource`, to describe valid random access iterator-like
+Let us define a named requirement, ``AdaptingIteratorSource``, to describe valid random access iterator-like
 types that can be used as source for oneDPL iterators as described below.
-The type :code:`Iter` satisfies the :code:`AdaptingIteratorSource` named requirement if it is any of the following:
+The type ``Iter`` satisfies the ``AdaptingIteratorSource`` named requirement if it is any of the following:
 
  * A random access iterator
- * The unspecified iterator-like type returned by :code:`oneapi::dpl::begin` or :code:`oneapi::dpl::end`
- * A :code:`permutation_iterator`
- * A :code:`transform_iterator`
- * A :code:`counting_iterator`
- * A :code:`discard_iterator`
- * A :code:`zip_iterator`
+ * The unspecified iterator-like type returned by ``oneapi::dpl::begin`` or ``oneapi::dpl::end``
+ * A ``permutation_iterator``
+ * A ``transform_iterator``
+ * A ``counting_iterator``
+ * A ``discard_iterator``
+ * A ``zip_iterator``
 
 .. code:: cpp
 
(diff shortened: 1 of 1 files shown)


source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -262,6 +262,18 @@ Iterators
 The oneDPL iterators are defined in the :code:`<oneapi/dpl/iterator>` header,
 in :code:`namespace oneapi::dpl`.
 
+Let us define a named requirement, :code:`ValidParallelIteratorSource`, to describe valid random access iterator-like
+types that can be used as source for oneDPL's iterators as described below.
+The type :code:`Iter` satisfies the :code:`ValidParallelIteratorSource` named requirement if it satisfies
+at least one of the following:
+ * :code:`Iter` satisfies the C++ named requirement :code:`LegacyRandomAccessIterator`
```
akukanov, line 269: Unfortunately, `LegacyRandomAccessIterator` used at cppreference.com is not a thing in the standard.

In C++17, random access iterator requirements do not have a special name. In C++20 and beyond, the named requirement is called `Cpp17RandomAccessIterator`.

danhoeflinger: I will take a renewed look at this in the context of `Cpp17RandomAccessIterator`.  
Assuming it is a good match for us still, is it reasonable to use `Cpp17RandomAccessIterator` within oneDPL's spec (given that oneDPL based upon C++17)?  
If not, should we try to distil down the key requirements we need and define our own baseline for a random access-like iterator?

danhoeflinger: We do use the term "random access iterators" when describing the types of data which may be passed to oneDPL algorithms.  Should we just use that term here or should we try to define it more explicitly?

akukanov: Let's just use "random access iterator" for now.

danhoeflinger: done.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -262,6 +262,19 @@ Iterators
 The oneDPL iterators are defined in the :code:`<oneapi/dpl/iterator>` header,
 in :code:`namespace oneapi::dpl`.
 
+Let us define a named requirement, :code:`ValidParallelIteratorSource`, to describe valid random access iterator-like
+types that can be used as source for oneDPL's iterators as described below.
```
akukanov, line 266:
```suggestion
types that can be used as source for oneDPL iterators as described below.
```

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -262,6 +262,19 @@ Iterators
...
+The type :code:`Iter` satisfies the :code:`ValidParallelIteratorSource` named requirement if it satisfies
+at least one of the following:
+
+ * :code:`Iter` is a random access iterator
+ * :code:`Iter` is the unspecified iterator-like type returned by :code:`oneapi::dpl::begin` or :code:`oneapi::dpl::end`
+ * :code:`Iter` is a valid :code:`permutation_iterator`
+ * :code:`Iter` is a valid :code:`transform_iterator`
+ * :code:`Iter` is a valid :code:`counting_iterator`
+ * :code:`Iter` is a valid :code:`discard_iterator`
+ * :code:`Iter` is a valid :code:`zip_iterator`
```
akukanov, line 276: What does "valid" indicate here? We are speaking about types, not instances.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -262,6 +262,19 @@ Iterators
 The oneDPL iterators are defined in the :code:`<oneapi/dpl/iterator>` header,
 in :code:`namespace oneapi::dpl`.
 
+Let us define a named requirement, :code:`ValidParallelIteratorSource`, to describe valid random access iterator-like
+types that can be used as source for oneDPL's iterators as described below.
+The type :code:`Iter` satisfies the :code:`ValidParallelIteratorSource` named requirement if it satisfies
+at least one of the following:
```
akukanov, line 268: Change "at least one" to "any"? Perhaps not much difference, but "at least one" confuses me a bit because the items in the list are seemingly exclusive.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -262,6 +262,19 @@ Iterators
 The oneDPL iterators are defined in the :code:`<oneapi/dpl/iterator>` header,
 in :code:`namespace oneapi::dpl`.
 
+Let us define a named requirement, :code:`ValidParallelIteratorSource`, to describe valid random access iterator-like
```
akukanov, line 265: I think the word "valid" is redundant in the name; it really adds nothing to the meaning. And I do not know how to interpret "ParallelIterator".

I would suggest something like `AlteringIteratorSource` or maybe `Adjusting` or `Adapting` (or whatever English adjective describes best what transform. permutation and zip iterators do with their sources).

source/elements/oneDPL/source/parallel_api.rst
danhoeflinger: "any" is fine.  I think that `counting_iterator` also satisfies "random access iterator" for instance. The overlap is generally only between "random access iterator" and other types.

source/elements/oneDPL/source/parallel_api.rst
danhoeflinger: My intention is to indicate that if you have something like `zip_iterator<transform_iterator<bad_source, func>,counting_iterator<int>>` , where `bad_source` is does not meet the requirements that `transform_iterator` lays out, that `bad_source` iterator ruins all combinations of types based upon it.  I guess that is obvious without any word like "valid", so we can remove it.

source/elements/oneDPL/source/parallel_api.rst
danhoeflinger: I can think about this, and propose a better name.

source/elements/oneDPL/source/parallel_api.rst
danhoeflinger: applied.

source/elements/oneDPL/source/parallel_api.rst
danhoeflinger: rephrased from your suggestion to remove valid.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -391,7 +404,19 @@ defined by the source iterator provided, and whose iteration order over the dere
...
+
+The type :code:`IndexMap` must satisfy at least one of the following:
+
+ * :code:`IndexMap` is a random access iterator
+ * :code:`IndexMap` is the unspecified iterator-like type returned by :code:`oneapi::dpl::begin` or :code:`oneapi::dpl::end`
+ * :code:`IndexMap` is a valid :code:`permutation_iterator`
+ * :code:`IndexMap` is a valid :code:`transform_iterator`
+ * :code:`IndexMap` is a valid :code:`counting_iterator`
+ * :code:`IndexMap` is a functor with a signature equivalent to :code:`T operator()(const T&) const` where
+   :code:`T` is a `std::iterator_traits<SourceIterator>::difference_type`
```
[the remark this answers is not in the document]
danhoeflinger: done.

source/elements/oneDPL/source/parallel_api.rst
akukanov: For something like you described, the requirement is violated for `transform_iterator<bad_source, func>` - which I would assume lead to a compilation error because the type cannot be instantiated.
The situation is different though if such a type can be instantiated but the behavior is wrong/undefined. Is that the case?

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -262,6 +262,18 @@ Iterators
 The oneDPL iterators are defined in the :code:`<oneapi/dpl/iterator>` header,
 in :code:`namespace oneapi::dpl`.
 
+Let us define a named requirement, :code:`AdaptingIteratorSource`, to describe valid random access iterator-like
+types that can be used as source for oneDPL iterators as described below.
+The type :code:`Iter` satisfies the :code:`AdaptingIteratorSource` named requirement if is any of the following:
```
akukanov, line 267:
```suggestion
The type :code:`Iter` satisfies the :code:`AdaptingIteratorSource` named requirement if it is any of the following:
```

source/elements/oneDPL/source/parallel_api.rst
danhoeflinger: Yes, I believe we could make a source type `bad_source` which would not cause compilation errors but would not satisfy the requirements set out there and the behavior would therefore be undefined.  I need to think a little to contrive such a type, but I think it exists.  

Also, there may be compile errors for `bad_source` iterators which may not be so clear, as we do not currently have some compile time check for this listing of types at the top level with a clear user facing error. It may manifest as some missing operator buried within an algorithm.

akukanov: The types of errors do not matter as it comes to the specification; it's a quality of implementation issue. However if it is not always a compilation error, that changes the things.

I see two possible approaches:
- allow this in the specification, so that it matches the current implementation; or
- disallow this in the specification (i.e. require a compile-time error and not UB), and make a "known issue" note for the implementation that it violates the spec - of course with the intention to fix it later.

My preference goes to the second option.

danhoeflinger: Let me experiment with this a little and make sure I am not wrong.  It may be that it would result in compiler errors or correct (but unspecified) results.  For instance, I think we don't need ALL the requirements of random access iterator in practice, but it covers what we do need.

I'll try to come up with some example(s) which result in a successful build, but incorrect results by not meeting this specification.

danhoeflinger: I have not yet tested this in practice, but lets consider an iterator type `bad_source` which is not a "random access iterator" as the c++20 standard defines it. Specifically, lets say it is defined such that the bracket operator `iter[n]` does not return something equivalent to `*(iter + n)`, but rather it just ignores `n` and returns `*iter`.  

`bad_source` still defines all the operators, functions, etc. which would be needed by a random access iterator, it just does not match the semantics required of a "random access iterator" for `operator[]`.  If this were used as input to a oneDPL algorithm, it would cause problems at runtime, but I think it would build successfully. Similarly, if inputs of `transform_iterator<bad_source, func>`, or `zip_iterator<transform_iterator<bad_source, func>, int*>`  were used as input to a oneDPL algorithm they would experience similar issues at runtime, not compile time I believe. 

I think that this is OK, because they are disobeying the spec, so their behavior is undefined.  `bad_source` is not a valid oneDPL input, nor are iterators which wrap `bad_source`.  The original point was merely that our iterators can be used as source iterators for our iterators where specified, but only when they adhere to the specification of those iterators. 

Do we need extra language to state this, or do you think this is covered by the current language?

akukanov: OK, I see now what you mean. It's about the source violating semantical requirements that cannot be checked at compile time, and so our adjusting iterators "inherit" the undefined behavior of the source. I do not think any special language is necessary for this.

Thinking of a case where our iterators can **introduce** undefined behavior even though the source was fully conformant, I find the `permutation_iterator` standing out, as both its data source and the map might be perfectly valid, but using the map values to access the data source might make that access invalid. This case may deserve some extra wording indicating a potential undefined behavior if certain conditions are not met.

danhoeflinger: I believe that the requirements specified here for the "source data" and for the "map" of a `permutation_iterator` allow it to function appropriately and not introduce any undefined behavior as long as the requirements are met.  I will take another look through this to confirm.

We explored this in a good amount of detail with our input_data_sweep tests, and a number of bugfixes earlier this year.

danhoeflinger: In our existing implementation of oneDPL and permutation iterator, there is an issue when using a `std::vector::iterator` (host iterator) as the source iterator for a permutation iterator.  This is because without sweeping through the whole map vector, we do not know what range of data to prepare by copying into a buffer.  Here in the spec, we don't specify the details of getting data to the device (host iterators, USM allocators, etc.), so this is not something that makes sense to be in the specification.  Otherwise, the `permutation_iterator` should be reliable as described here without introducing any undefined behavior.

This is noted in the oneDPL implementation's documentation [here](https://github.com/oneapi-src/oneDPL/blob/1266c3867ab2aa234519b8f7cffae3e95a2483bc/documentation/library_guide/parallel_api/iterators.rst?plain=1#L105).

akukanov: Thank you. This discussion is resolved then.

akukanov (approved):

PR #552, opened 2024-06-27
in-place conditions for ccl

rscohn2:


garzaran (approved):
This looks good. Thanks so much!

PR #555, opened 2024-07-13
Added blocked_rangeNd

tobiasweinzierl80:
An n-dimensional range has been in Intel's reference implementation for quite a while. It is an absolutely essential feature for a lot of scientific codes and hence should be part of the spec. This proposal starts from Intel's reference implementation but adds an array-based constructor which we found extremely useful in our codes. A PR to the reference implementation is submitted.

### DIFF
--- a/source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
+++ b/source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
@@ -1,0 +1,153 @@
+.. SPDX-FileCopyrightText: 2019-2021 Intel Corporation
+..
+.. SPDX-License-Identifier: CC-BY-4.0
+
+===============
+blocked_range3d
+===============
+**[algorithms.blocked_range3d]**
+
+Class template that represents a recursively divisible N-dimensional half-open interval.
+
+A ``blocked_rangeNd`` is the N-dimensional extension of ``blocked_range``, ``blocked_range2d`` and ``blocked_range3d``.
+We can interpret it as a tensor-product of N instances of ``blocked_range``.
+Different to ``blocked_range2d`` and ``blocked_range3d``, all ranges have to be specified over the same type Value.
+Therefore, blocked_rangeNd<int,2> is a specialisation of blocked_range2d where both dimensions have int type.
+
+.. code:: cpp
+
+    namespace oneapi {
+        namespace tbb {
+            template<typename Value, unsigned int N, typename = detail::make_index_sequence<N>>
+            class blocked_rangeNd {
+            public:
+                // Types
+                using value_type = Value;
+                using dim_range_type = tbb::blocked_range<Value>;
+
+                // Constructors
+                blocked_rangeNd(const dim_range_type<Is>&... args);
+                blocked_rangeNd(
+                  Value begin[N],
+                  Value end[N],
+                  typename tbb::blocked_range<value_type>::size_type grainsize
+                );
+                blocked_rangeNd(blocked_rangeNd& r, proportional_split proportion); 
+                blocked_rangeNd(blocked_rangeNd& r, split proportion); 
+               
+                // Capacity
+                bool empty() const;
+
+                // Access
+                bool is_divisible() const;
+                static constexpr unsigned int ndims();
+                const tbb::blocked_range<value_type>& dim(unsigned int dimension) const;
+            };
+
+        } // namespace tbb
+    } // namespace oneapi        
+
+Requirements:
+
+* The *Value* must meet the :doc:`blocked_range requirements <../../named_requirements/algorithms/blocked_range_val>`
+
+Member types
+------------
+
+.. code:: cpp
+
+    using value_type = Value;
+
+The type of the page values.
+
+.. code:: cpp
+
+    using page_range_type = blocked_range<PageValue>;
+
+The type of the values in one direction out of the N dimensions.
+
+Member functions
+----------------
+
+.. code:: cpp
+
+    blocked_rangeNd(const dim_range_type<Is>&... args);
+
+**Effects:**  Constructs a ``blocked_rangeNd`` representing an N-dimensional space of values.
+The space is the half-open Cartesian product of ranges each having its own grain size.
+The number of args has to match N.
+
+
+.. code:: cpp
+
+    blocked_rangeNd(
+      Value begin[N],
+      Value end[N],
+      typename tbb::blocked_range<value_type>::size_type grainsize
+    );
+
+**Effects:**  Constructs a ``blocked_rangeNd`` representing an N-dimensional space of values.
+The space is the half-open Cartesian product of ranges with ``[begin[0], end[0]) x [begin[1], end[1]) x ...`` each having the same grain size.
+
+**Example:**  The statement ``blocked_rangeNd<int,4> r( {1,2,3,4}, {5,6,7,8}, 4 );`` constructs a four-dimensional
+space that contains all value pairs of the form ``(i, j, k, l)``, where ``i`` ranges from 1 to 5 with a grain size of 1,
+``j`` ranges from 2 to 6 with a grain size of 1, and so forth.
+    
+
+Basic splitting constructor.
+
+.. code:: cpp
+
+    blocked_rangeNd( blocked_rangeNd& range, split );
+
+Basic splitting constructor.
+
+**Requirements**: ``is_divisible()`` is true.
+
+**Effects**: Partitions ``range`` into two subranges. The newly constructed ``blocked_rangeNd`` is approximately
+the half of the original ``range``, and ``range`` is updated to be the remainder.
+Each subrange has the same grain size as the original ``range``. Splitting is done in any dimension.
+The choice of which axis to split is intended to cause, after repeated splitting, 
+subranges of approximately square/cubic/hypercubic shape if all grain sizes are the same.
+
+.. code:: cpp
+
+    blocked_rangeNd( blocked_range3d& range, proportional_split proportion );
+
+Proportional splitting constructor.
+
+**Requirements**: ``is_divisible()`` is true.
+
+**Effects**: Partitions ``range`` into two subranges in the given ``proportion``
+across one of its axes. The choice of which axis to split is made in the same way as for the basic splitting
+constructor; then, proportional splitting is done for the chosen axis. The second axis and the grain sizes for
+each subrange remain the same as in the original range.
+
+.. code:: cpp
+
+    bool empty() const;
+
+**Effects**: Determines if range is empty.
+
+**Returns:** ``pages.empty()||rows().empty()||cols().empty()``
+
+.. code:: cpp
+
+    bool is_divisible() const;
+
+**Effects**: Determines if the range can be split into subranges.
+
+**Returns:** any dim(int) returns is_divisible().
+
+.. code:: cpp
+
+    const page_range_type& dim(int) const;
+
+**Returns:**  Range containing the range of the value space along axis specified by argument.
+
+See also:
+
+* :doc:`blocked_range <blocked_range_cls>`
+* :doc:`blocked_range2d <blocked_range2d_cls>`
+* :doc:`blocked_range3d <blocked_range3d_cls>`
+
--- a/source/elements/oneTBB/source/algorithms.rst
+++ b/source/elements/oneTBB/source/algorithms.rst
@@ -39,6 +39,7 @@ Types that meet the :doc:`Range requirements <named_requirements/algorithms/rang
    algorithms/blocked_ranges/blocked_range_cls.rst
    algorithms/blocked_ranges/blocked_range2d_cls.rst
    algorithms/blocked_ranges/blocked_range3d_cls.rst
+   algorithms/blocked_ranges/blocked_rangeNd_cls.rst
 
 .. _Partitioners:
 


source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,153 @@
...
+            class blocked_rangeNd {
+            public:
+                // Types
+                using value_type = Value;
+                using dim_range_type = tbb::blocked_range<Value>;
+
+                // Constructors
+                blocked_rangeNd(const dim_range_type<Is>&... args);
+                blocked_rangeNd(
+                  Value begin[N],
```
vossmjp, line 31: Is there a reason to prefer blocked_rangeNd<int,4> r( {1,2,3,4}, {5,6,7,8}, 4 ) over  blocked_rangeNd<int,4> r( {1, 5}, {2, 6}, {3, 7}, {4, 8}, 4 )?

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,153 @@
...
+
+.. code:: cpp
+
+    using value_type = Value;
+
+The type of the page values.
+
+.. code:: cpp
+
+    using page_range_type = blocked_range<PageValue>;
```
vossmjp, line 65:
```suggestion
    using dim_range_type = blocked_range<Value>;
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,153 @@
...
+
+The type of the values in one direction out of the N dimensions.
```
vossmjp, line 67:
```suggestion
The type of the values in one dimension out of the N dimensions.
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,153 @@
...
+
+**Effects:**  Constructs a ``blocked_rangeNd`` representing an N-dimensional space of values.
+The space is the half-open Cartesian product of ranges with ``[begin[0], end[0]) x [begin[1], end[1]) x ...`` each having the same grain size.
+
+**Example:**  The statement ``blocked_rangeNd<int,4> r( {1,2,3,4}, {5,6,7,8}, 4 );`` constructs a four-dimensional
+space that contains all value pairs of the form ``(i, j, k, l)``, where ``i`` ranges from 1 to 5 with a grain size of 1,
+``j`` ranges from 2 to 6 with a grain size of 1, and so forth.
+    
+
+Basic splitting constructor.
```
vossmjp, line 97:
```suggestion
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,153 @@
...
+constructor; then, proportional splitting is done for the chosen axis. The second axis and the grain sizes for
+each subrange remain the same as in the original range.
+
+.. code:: cpp
+
+    bool empty() const;
+
+**Effects**: Determines if range is empty.
+
+**Returns:** ``pages.empty()||rows().empty()||cols().empty()``
```
vossmjp, line 132: Seems like this reuses blocked_range_3d description, but there are no pages, rows or cols.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,153 @@
...
+
+    bool is_divisible() const;
+
+**Effects**: Determines if the range can be split into subranges.
+
+**Returns:** any dim(int) returns is_divisible().
+
+.. code:: cpp
+
+    const page_range_type& dim(int) const;
```
vossmjp, line 144:
```suggestion
    const dim_range_type& dim(int) const;
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,153 @@
...
+
+**Returns:**  Range containing the range of the value space along axis specified by argument.
```
vossmjp, line 146: Should we say "dimension" instead of "axis", just to reduce terminology.

vossmjp (commented):
Mostly suggesting fixes for typos, but also a couple of additional comments.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,153 @@
+.. SPDX-FileCopyrightText: 2019-2024 Intel Corporation
+..
+.. SPDX-License-Identifier: CC-BY-4.0
+
+===============
+blocked_range3d
```
vossmjp, line 6:
```suggestion
blocked_rangeNd
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,153 @@
...
+A ``blocked_rangeNd`` is the N-dimensional extension of ``blocked_range``, ``blocked_range2d`` and ``blocked_range3d``.
+We can interpret it as a tensor-product of N instances of ``blocked_range``.
+Different to ``blocked_range2d`` and ``blocked_range3d``, all ranges have to be specified over the same type Value.
+Therefore, blocked_rangeNd<int,2> is a specialisation of blocked_range2d where both dimensions have int type.
+
+.. code:: cpp
+
+    namespace oneapi {
+        namespace tbb {
+            template<typename Value, unsigned int N, typename = detail::make_index_sequence<N>>
```
vossmjp, line 21:
```suggestion
            template<typename Value, unsigned int N>
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,153 @@
...
+            class blocked_rangeNd {
+            public:
+                // Types
+                using value_type = Value;
+                using dim_range_type = tbb::blocked_range<Value>;
```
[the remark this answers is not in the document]
akukanov: I think it depends on whether this helper is necessary to describe the constructor that takes exactly N blocked_range objects. I'm thinking of how to do it best.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
akukanov: Note that the constructor at line 29 can also work with braced init lists, like this: `blocked_rangeNd<int,4> r{ {1,5,4}, {2,6,4}, {3,7,4}, {4,8,4}};` if the grainsize is explicitly specified as 4.

akukanov: I have looked at the use of blocked_rangeNd in Peano (https://gitlab.lrz.de/hpcsoftware/Peano/-/blob/p4/src/peano4/utils/tbb/Loop.h), and I think we can define two constructors from arrays:
```
blocked_rangeNd( Value begin[N], Value end[N], size_type grainsize[N] );
blocked_rangeNd( Value size[N], size_type grainsize = 1 ); // assumes all dimensions are {0, size[i]}
```
The first one is the alternative way to specify dimensions in "structure of arrays"-like manner, and the second is the convenient helper to both set identical grainsize **and** set all dimensions to start from 0. This one is what Peano would use.

While constructors that do only one of the two simplifications are also possible, I am not sure if they are worth having. Also we can add more syntax sugar later if justification appears.

What do you think?

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,151 @@
...
+                );
+                blocked_rangeNd(blocked_rangeNd& r, proportional_split proportion); 
+                blocked_rangeNd(blocked_rangeNd& r, split proportion); 
+               
+                // Capacity
+                bool empty() const;
+
+                // Access
+                bool is_divisible() const;
+                static constexpr unsigned int ndims();
```
akukanov, line 43: This static member function seems not described anywhere.

Also I would think of a better name; this one seems too cryptic. Maybe `dim_count` or even `dimension_count`?

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,151 @@
...
+                const tbb::blocked_range<value_type>& dim(unsigned int dimension) const;
+            };
+
+        } // namespace tbb
+    } // namespace oneapi        
+
+Requirements:
+
+* The *Value* must meet the :doc:`blocked_range requirements <../../named_requirements/algorithms/blocked_range_val>`
```
akukanov, line 52:
```suggestion
* The *Value* must meet the :doc:`BlockedRangeValue requirements <../../named_requirements/algorithms/blocked_range_val>`
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
sasalla23: I will go with `dim_count`, as `dim` is already used in place of `dimension` in several other instances in the context of blocked_rangeNd.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
vossmjp: I think the first doesn't offer much over the braced init lists already supported by the constructor at line 29. But then I prefer the begin & end to be near each other for readability. The second suggested constructor seems very reasonable since its I'd suppose that it is the most common case.

akukanov: It's OK to go with only the most simplified version that omits begins and uses a single grainsize. Having a general way to construct an ND range from arrays (as an alternative to a set of 1D ranges) might in principle be meaningful, but if you think it doesn't add much value, I'm fine without it.

tobiasweinzierl80: I think that's reasonable. I honestly never saw a real case for different grain sizes in the different dimensions unless(!) the innermost where you wanna vectorise. But then I'm not sure how good the compiler supports that kind of multilevel parallelisation (vectorisation inside, shared mem outside) anyway. So let's stick with one grain size instead of a set.

sasalla23: I implemented what I have understood to be the consensus here in my new commit: The two main constructors whould be `blocked_rangeNd(const dim_range_type<Is>&... args);` and `blocked_rangeNd( Value size[N], size_type grainsize = 1 );`. Please point out in case I misunderstood the discussion and the original `blocked_rangeNd( Value begin[N], Value end[N], size_type grainsize = 1 );` is still needed.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,156 @@
...
+                using dim_range_type = tbb::blocked_range<Value>;
+
+                // Constructors
+                blocked_rangeNd(const dim_range_type<Is>&... args);
+                blocked_rangeNd(
+                  Value size[N], 
+                  typename tbb::blocked_range<value_type>::size_type grainsize = 1
+                );
+                blocked_rangeNd(blocked_rangeNd& r, proportional_split proportion); 
+                blocked_rangeNd(blocked_rangeNd& r, split proportion);
```
vossmjp, line 36:
```suggestion
                blocked_rangeNd(blocked_rangeNd& r, split); 
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,156 @@
...
+
+**Effects**: Determines if the range can be split into subranges.
+
+**Returns:** any dim(int) returns is_divisible().
+
+.. code:: cpp
+
+   static constexpr unsigned int dim_count();
+
+**Returns:** Number of dimensions specified by the template argument ``N``.
```
vossmjp, line 143:
```suggestion
**Returns:** Number of dimensions, as specified by the class template argument ``N``.
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
akukanov: > In the suggested version grainsize is set to 1 by default. Is this the desired way?

It matches the default grainsize value for all other blocked ranges.

akukanov:
I have added `using size_type = dim_range_type::size_type;` to the class definition, and changed some of the methods to use it. It needs to be added to the implementation.

akukanov:
In the last commit, I tried to specify the constructor from N blocked ranges in a different way which avoids the use of an unspecified parameter pack.

@vossmjp please take a look. If this does not seem good, the only other option I could think of is uncovering the implementation trick we use.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,167 @@
...
+            class blocked_rangeNd {
+            public:
+                // Types
+                using value_type = Value;
+                using dim_range_type = blocked_range<value_type>;
+                using size_type = typename dim_range_type::size_type;
+
+                // Constructors
+                blocked_rangeNd(const dim_range_type& dim0 /*, ... - exactly N parameters of the same type*/);
+                blocked_rangeNd(value_type size[N], size_type grainsize = 1);
```
akukanov, line 32: As far as I know, the parameter of an array type will not match to a braced-init-list argument, so `blocked_rangeNd<...>(/*sizes*/{100,200}, /*grainsize*/10)` will not compile. If we want it to compile, we'd need to change the parameter to e.g. `std::array<value_type,N>` or add another constructor.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,171 @@
...
+.. SPDX-FileCopyrightText: Contributors to the oneAPI Specification project.
+..
+.. SPDX-License-Identifier: CC-BY-4.0
+
+===============
+blocked_rangeNd
+===============
+**[algorithms.blocked_rangeNd]**
+
+Class template that represents a recursively divisible N-dimensional half-open interval.
```
aepanchi, line 11:
```suggestion
The class template that represents a recursively divisible N-dimensional half-open interval.
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,171 @@
...
+
+A ``blocked_rangeNd`` is the N-dimensional extension of ``blocked_range``, ``blocked_range2d`` and ``blocked_range3d``.
```
aepanchi, line 13:
```suggestion
A ``blocked_rangeNd`` is the N-dimensional extension of ``blocked_range``, ``blocked_range2d``, and ``blocked_range3d``.
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,171 @@
...
+    blocked_rangeNd( const dim_range_type& dim0 /*, ... - exactly N parameters of the same type*/ );
+
+**Effects:**  Constructs a ``blocked_rangeNd`` representing an N-dimensional space of values.
+The space is the half-open Cartesian product of ranges ``dim0 x ...`` each having its own grain size.
+The constructor must take exactly N arguments which types match to ``const dim_range_type&``.
+
+**Example:** For ``blocked_rangeNd<int,4>``, this constructor is equivalent to
+``blocked_rangeNd( const blocked_range<int>&, const blocked_range<int>&, const blocked_range<int>&, const blocked_range<int>& )``.
+
+.. note::
```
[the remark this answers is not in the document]
aepanchi: Maybe rephrase to smth like it for an easier understanding: 

A variadic template constructor, template <typename... Dims> blocked_rangeNd(const Dims&... dims), does not work correctly in all cases. Even if you set specific size and type rules for the Dims parameter pack, it does not fully comply because the types in Dims cannot be automatically determined when using braced initialization lists. As a result, expressions like blocked_rangeNd<int, 4>{{0,1},{0,2},{0,3},{0,4}} cause a compilation error.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,171 @@
...
+.. code:: cpp
+
+    blocked_rangeNd( value_type size[N], size_type grainsize = 1 );
+
+**Effects:**  Constructs a ``blocked_rangeNd`` representing an N-dimensional space of values.
+The space is the half-open Cartesian product of ranges ``[0, size[0]) x [0, size[1]) x ...``
+each having the same grain size.
+
+**Example:**  The statement ``blocked_rangeNd<int,4> r( {5,6,7,8}, 4 );`` constructs a four-dimensional
+space that contains all value pairs of the form ``(i, j, k, l)``, where ``i`` ranges from 0 (included)
```
aepanchi, line 104:
```suggestion
space that contains all value pairs of the ``(i, j, k, l)`` form, where ``i`` ranges from 0 (included)
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,171 @@
...
+.. code:: cpp
+
+    blocked_rangeNd( blocked_rangeNd& range, proportional_split proportion );
+
+Proportional splitting constructor.
+
+**Requirements**: ``is_divisible()`` is true.
+
+**Effects**: Partitions ``range`` into two subranges in the given ``proportion`` across one of its dimensions.
+The choice of which dimenstion to split is made in the same way as for the basic splitting constructor;
```
aepanchi, line 132:
```suggestion
The choice of which dimension to split is the same as for the basic splitting constructor. 
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,171 @@
...
+then, proportional splitting is done for the chosen one. Other dimensions and the grain sizes for
```
aepanchi, line 133:
```suggestion
The proportional splitting is done for the selected one. Other dimensions and the grain sizes for
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,171 @@
...
+
+**Returns:** The number of dimensions, as specified by the class template argument ``N``.
+
+.. code:: cpp
+
+    bool empty() const;
+
+**Effects**: Determines if range is empty.
+
+**Returns:** True if any of the range dimensions is empty; false, otherwise.
```
aepanchi, line 148:
```suggestion
**Returns:** ``True`` if any range dimensions is empty; false, otherwise.
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,171 @@
...
+
+**Returns:** True if for any of the range dimensions ``is_divisible()`` is true; false, otherwise.
+
+.. code:: cpp
+
+    const dim_range_type& dim(unsigned int dimension) const;
+
+**Requirements**: 0 <= ``dimension`` < N.
+
+**Returns:**  ``blocked_range`` containing the value space along the dimension specified by the argument.
```
aepanchi, line 164:
```suggestion
**Returns:** ``blocked_range`` containing the value space along the dimension specified by the argument.
```

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
akukanov: Won't apply; consistent with other pages.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,171 @@
...
+                bool is_divisible() const;
+                const dim_range_type& dim(unsigned int dimension) const;
+            };
+
+        } // namespace tbb
+    } // namespace oneapi        
+
+Requirements:
+
+* The *Value* must meet the :doc:`BlockedRangeValue requirements <../../named_requirements/algorithms/blocked_range_val>`
```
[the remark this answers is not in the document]
akukanov: Won't apply; consistent with other pages.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,171 @@
...
+
+Member types
```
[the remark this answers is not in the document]
akukanov: Won't apply; consistent with other pages.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,171 @@
...
+.. code:: cpp
+
+    blocked_rangeNd( blocked_rangeNd& range, split );
+
+Basic splitting constructor.
+
+**Requirements**: ``is_divisible()`` is true.
+
+**Effects**: Partitions ``range`` into two subranges. The newly constructed ``blocked_rangeNd`` is approximately
+the half of the original ``range``, and ``range`` is updated to be the remainder.
```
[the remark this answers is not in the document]
akukanov: In this context, the definite article seems more appropriate.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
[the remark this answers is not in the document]
akukanov: This note is for class implementors, not for its users. It is intended to describe a potential pitfall of a seemingly equivalent but in fact subtly different form of the class constructor, which therefore should not be used by implementors as a substitution for the constructor specified above the note. This is why I use Subjunctive mood in the note instead of simpler verb forms.

The use of a note (and not a plain text) is also intentional, as I wanted to indicate that this part is informational, not normative (see more details at https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:nonnormativerefs). Perhaps this approach is not used consistently in the specification, but I find it applicable and quite useful here. I guess this note can be changed to a caution - I see some uses of it in the oneTBB spec.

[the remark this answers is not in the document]
akukanov: Reworked it somewhat, hopefully for better.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
akukanov: Reworked.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,175 @@
...
+
+.. code:: cpp
+
+    blocked_rangeNd( value_type size[N], size_type grainsize = 1 );
+
+**Effects:**  Constructs a ``blocked_rangeNd`` representing an N-dimensional space of values.
+The space is the half-open Cartesian product of ranges ``[0, size[0]) x [0, size[1]) x ...``
+each having the same grain size.
+
+**Example:**  The ``blocked_rangeNd<int,4> r( {5,6,7,8}, 4 );`` statement constructs a four-dimensional
```
pavelkumbrasev, line 105: All the other ranges constructors gives you ability to specify begins of the ranges. 
At least from my perspective it less confusing to have both begins and ends explicitly specified.
Otherwise, `blocked_rangeNd<int,2> r( {1,6} );` could be confused into `blocked_rangeNd<int,2> r( {/*begin*/1,/*end*/6} );`.

akukanov: We discussed it here https://github.com/uxlfoundation/oneAPI-spec/pull/555#discussion_r1730432024 and agreed to only provide the "simplest" version that takes the array of sizes.

> Otherwise, `blocked_rangeNd<int,2> r( {1,6} );` could be confused into `blocked_rangeNd<int,2> r( {/*begin*/1,/*end*/6} );`.

`blocked_rangeNd<int,2> r( {/*begin*/1,/*end*/6} );.` should have a second dimension though to be meaningful, no?

pavelkumbrasev: >should have a second dimension though to be meaningful, no?

Also true but still looks a bit confusing 😄

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
akukanov: It seems I was wrong. The problem is not with the array type vs std::array, but with the by-value parameter declaration.
And the following should work: `blocked_rangeNd(const value_type (&size)[N], size_type grainsize = 1);`

See https://godbolt.org/z/Maer5Kecs for some prototyping.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,175 @@
...
+            class blocked_rangeNd {
+            public:
+                // Types
+                using value_type = Value;
+                using dim_range_type = blocked_range<value_type>;
+                using size_type = typename dim_range_type::size_type;
+
+                // Constructors
+                blocked_rangeNd(const dim_range_type& dim0 /*, ... - exactly N parameters of the same type*/);
+                blocked_rangeNd(const value_type (&size)[N], size_type grainsize = 1);
```
akukanov, line 33: Do we want to address support for class template argument deduction in this PR as well, or better separately? I guess it might be non-trivial and require some experiments first.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,175 @@
...
+Unlike ``blocked_range2d`` and ``blocked_range3d``, all ranges must be specified over the same ``Value`` type.
+Therefore, ``blocked_rangeNd<int,2>`` is a specialization of ``blocked_range2d`` where both dimensions are ``int``.
+
+.. code:: cpp
+
+    namespace oneapi {
+        namespace tbb {
+
+            template<typename Value, unsigned int N>
+            class blocked_rangeNd {
```
rarutyun, line 24: should `blocked_rangeNd` be `blocked_range_nd`? I might understand why we named `blocked_range2d` and `3d` without underscore but for `_nd` we have a mix of `snake_case` and `camelCase` now

akukanov: Ah, the naming question! :)

To feed the discussion: SYCL has `nd_range` and `nd_item`, and C++23 has `mdspan` (but the real analogue to our class is called `extents` in C++).

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
```diff
@@ -0,0 +1,175 @@
...
+            public:
+                // Types
+                using value_type = Value;
+                using dim_range_type = blocked_range<value_type>;
```
akukanov, line 28: One more naming question: consider `dimension_type` as the name, for clarity and also to avoid interpretation of "dim range type" as "shadowy" etc. Or does it seem important to highlight the range nature of the dimensions?

tobiasweinzierl: Agreed. However, this name is consistent with the 2d and 3d range if I remember correctly. So if we change it here, we become inconsistent with the other types. This all would not have been an issue, but a previous comment of colleagues from Intel suggested that we make it public. And then inconsistencies matter. My opinion.

akukanov: Thanks for pointing this out, @tobiasweinzierl. I agree that consistency matters. Closing this question.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
vossmjp: My preference is to be silent about supporting template argument deduction in this version of the spec even while developing an implementation that supports it. This would let us do the experiments etc. without committing to anything. The spec would not be incompatible with an implementation whether the implementation has support or not.

akukanov: That's OK for me. I will keep the comment open for some time, in case others want to comment on this proposed resolution.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
akukanov: One more question to consider: should we also change "nd" to "md", following the pattern of C++ `mdspan`? This seems a bit more than purely cosmetic, as "md" stands for "multidimensional". And maybe even change the order of parts to `blocked_md_range`.

vossmjp: I don't have a strong preference on a specific name. Here are my thoughts:

-  I definitely do NOT like `blocked_rangend`since that is indecipherable
-  and I agree that `blocked_rangeNd` is not great either because it appears to be camelCase unlike any other name. 

So I'm left with `blocked_range_nd`, `blocked_nd_range` and `blocked_md_range`.  Of these, I prefer the options with "nd" or "md" in the middle to differentiate even more from the existing blocked ranges. I have the slightest preference for "nd" over "md" simply because it doesn't imply multiple dimensions; it could be 1. "n" doesn't imply that it has to be "multi". But I wouldn't object to "md" if others have a preference for it.

rarutyun: My preference is not `blocked_rangeNd` :)

`blocked_md_range`, `blocked_range_md`, `blocked_nd_range`, `blocked_range_nd` are all fine with me, with a different (but slight) level of preference.

I slightly prefer `md` to `nd`. Whether this `md` goes to an end to to a middle, I don't actually care. English-wise, perhaps `blocked_md_range` is better. Especially if we add aliases `blocked_2d_range` and `blocked_3d_range`. There is a caveat with those aliases, though... While the CTAD would work for `blocked_range_2d` and `_3d` since C++17, it would work for the respective aliases since C++20, which (IMO) looks weird from the user standpoint.

akukanov: > Hmm, I would prefer blocked_nd_range as it is a proper English word then. However we have blocked_range2d and blocked_range3d, so blocked_rangeNd is more consistent!

Looks like most if not all of us are OK with some inconsistency for the sake of a better name. We can think of `blocked_nd_range` as a new branch of `blocked_range` evolution, not a "descendant" of 2d/3d ranges (even though it borrows from those).

tobiasweinzierl: I can life with that as long as this rationale is documented here explicitly, i.e. we explain it in the documentation.

akukanov: That's reasonable, but if you do not mind I would add it to the oneTBB Developer Guide instead of the specification. Perhaps here: https://oneapi-src.github.io/oneTBB/main/tbb_userguide/Advanced_Topic_Other_Kinds_of_Iteration_Spaces.html

akukanov: > I can life with that as long as this rationale is documented here explicitly, i.e. we explain it in the documentation.

Addressed by https://github.com/uxlfoundation/oneAPI-spec/pull/555/commits/ed26d0c785a828daf34a28c9249b079ba1c0856c. Also I have opened https://github.com/oneapi-src/oneTBB/issues/1529 issue for updating the oneTBB developer guide.

source/elements/oneTBB/source/algorithms/blocked_ranges/blocked_rangeNd_cls.rst
akukanov: @pavelkumbrasev should we consider this conversation "resolved", or do you view it as an objection that needs to be addressed in the design?

pavelkumbrasev: "Resolved"

vossmjp (approved):
Looks good to me.

pavelkumbrasev (approved):
LGTM!

tobiasweinzierl80:
I think this is now in great shape.

PR #557, opened 2024-07-26
[oneMKL][RNG] Extend parameter type for random distributions

andreyfe1:
It was found that it's not necessary to make implementors use `std::vector` as a parameter type for `gaussian_mv`, `multinomial`, `poisson_v` distributions. They can also use `std::array` or `std::view` for example to pass data to distribution. This PR relaxes requirement for such parameter types.

Feel free to share your feedback


aelizaro (approved):
Thank you @andreyfe1, I like this change!

andreyfe1:
Please, provide your feedback till August 22. After that PR will be merged if there are no concerns

PR #559, opened 2024-07-29
[oneMKL][RNG] Add 2 distributions to Device API: beta, gamma.

andreyfe1:
Let's add new distributions to Device API: [beta](https://en.wikipedia.org/wiki/Beta_distribution), [gamma](https://en.wikipedia.org/wiki/Gamma_distribution).


andreyfe1:
Please, provide your feedback till August 22. After that PR will be merged if there are no concerns

Issue #551, opened 2024-06-27
Incomplete documentation about spec versioning


vmalia:
# Description
We recently had a customer request to provide APIs in Intel(R) oneMKL, for providing the version of the oneAPI spec that the product is compliant with.

During this investigation, I found that the oneAPI spec repository uses https://github.com/uxlfoundation/oneAPI-spec/blob/main/oneapi-doc.json file to define a version `string`, but the specification itself does not have any guidelines about `when` to rename the specification's major or minor version.

# Expected
1. The spec should define what a major and minor version update mean for the specification and the underlying product. Same for the `revision`.
2. The versioning scheme must be programmatically possible to implement, similar to semantic versioning.

rscohn2:
Here is the documentation: https://github.com/uxlfoundation/oneAPI-spec/blob/main/doc/versioning.rst

vmalia:
Thank you @rscohn2 . This addresses my first concern, but not the second.

PR #561, opened 2024-08-05
[RFC|oneMKL]Enable programmatic versioning

vmalia:
# Motivation
- Customers of oneMKL have requested a way to query the spec-compliance version of the implementation.
- The specification defines versioning but not in a programmatic way.
- Addresses #551 

# Inspiration
- SYCL versioning does not add the `revision` number:
--------
image
--------
- OpenMP specification itself keeps only `MAJOR.MINOR` version, with ongoing progress marked as `Technical Report <number>`. The `_OPENMP` macro provides a release date which maps onto a specification version, like this:
```c
#define _OPENMP 202111    //202111: OpenMP 5.2
```

# Proposal
- `<MAJOR><MINOR><REVISION>` such that MINOR and REVISION have two digits.
- Propose a macro for oneAPI and suggest a naming scheme for component macro names.


rscohn2:
@uxlfoundation/oneapi-spec-maintainers: Please review this versioning proposal

doc/versioning.rst
```diff
@@ -82,3 +82,14 @@ Example
...
   | oneAPI 1.1 rev 1
+
+Programmatic versioning
+=======================
+
+The specification version as mentioned above can be used programatically by appending all digits of the specification version in the format of <MAJOR><MINOR><REVISION>. MINOR and REVISION version use two digits. This versioning can be used to check the compatibility of the implementation with the specification version.
+
+Example
+  | oneAPI 1.1 rev 1 will be represented as a numerical value of 10101
+  | oneAPI 1.2 rev 1 will be represented as a numerical value of 10102
```
mkrainiuk, line 93: According to the proposed format shouldn't  `1.2 rev 1` be `10201` here instead of `10102` since `2` is a minor?

doc/versioning.rst
```diff
@@ -82,3 +82,14 @@ Example
   | oneAPI provisional 1.1 rev 2
   | oneAPI provisional 1.1 rev 3
   | oneAPI 1.1 rev 1
+
+Programmatic versioning
+=======================
+
+The specification version as mentioned above can be used programatically by appending all digits of the specification version in the format of <MAJOR><MINOR><REVISION>. MINOR and REVISION version use two digits. This versioning can be used to check the compatibility of the implementation with the specification version.
```
andrewtbarker, line 89: "appending all digits" to what? You are not proposing a macro/variable so I do not understand where these digits go.

andrewtbarker (commented):
I do not understand what is being proposed here.

Is the idea that each implementation of the spec gets to decide how to provide a version to the user (whether by a version string, a `#define`, an `int getSpecVersion()` function)? Then this change to the spec only specifies that the end of this string/macro/int has a certain representation?

doc/versioning.rst
vmalia: Fixed

doc/versioning.rst
vmalia: Added macro names.

doc/versioning.rst
```diff
@@ -82,3 +82,26 @@ Example
...
+
+Example
+  | oneAPI 1.1 rev 1 will be represented as a numerical value of 10101
+  | oneAPI 1.2 rev 1 will be represented as a numerical value of 10201
+  | oneAPI 1.2 rev 2 will be represented as a numerical value of 10202
+
+The implementation is required to define a macro with the numerical value of the specification version. The macro name should be `ONEAPI_SPEC_VERSION` and the value should be the numerical value of the specification version. The components within the specification, can implement their own macro with the numerical value of the specification version. The macro name should be `<COMPONENT>_SPEC_VERSION` and the value should be the numerical value of the specification version.
```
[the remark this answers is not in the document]
vmalia: > The PR title suggests that this change only affects oneMKL but it seems to be the scope of this change is broader that than.

I updated the title. You are right, it affects the full document and all components.

> What is the benefit of having ONEAPI_SPEC_VERSION  over only defining \<COMPONENT\>_SPEC_VERSION?

The spec currently does not support versioning on a component level, only oneAPI level, so the `<COMPONENT>_SPEC_VERSION` is optional, but `ONEAPI_SPEC_VERSION` is mandatory.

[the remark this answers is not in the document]
vmalia: Thanks, done.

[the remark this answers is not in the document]
akukanov: A mandatory macro with the same name in different projects is a terrible idea. On the user side, it can always happen that libraries implementing a different version of the specification are used, in which case one gets macro redefinition error.

vmalia:
> Is the idea that each implementation of the spec gets to decide how to provide a version to the user (whether by a version string, a `#define`, an `int getSpecVersion()` function)? Then this change to the spec only specifies that the end of this string/macro/int has a certain representation?

Yes. The goal of this RFC is to ensure that every implementation of the spec provides a numerically comparable value for its compliance with the spec version, through the proposed macros.

The implementation can additionally provide more ways to query that information, like APIs, but we do not mandate that in this proposal.

doc/versioning.rst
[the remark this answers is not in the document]
vmalia: Where should the changes be added then, so that they are reflected in the main oneAPI spec at https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.3-rev-1/ ?

doc/versioning.rst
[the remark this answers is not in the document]
Rbiessy: Thanks, I don't have the rights to resolve the comment so feel free to do it when I thumb up.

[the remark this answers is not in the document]
Rbiessy: Yes I also have concerns about defining the same macro in multiple projects.

doc/versioning.rst
```diff
@@ -37,11 +37,11 @@ Increment the:
    incompatible API changes, including removing APIs.
 
 2. MINOR version when adding minor functionality and API changes
-   that are backwards compatible.
+   that are backwards compatible. Supported values: 0-99 [1]
 
 3. REVISION when making backwards compatible bug fixes or any editing
    change in the document, including minor changes such as correcting
-   typos. Initial REVISION is 1.
+   typos. Initial REVISION is 1. Supported values: 1-99 [1]
```
akukanov, line 44: Is it even necessary to track revisions (in other words, editorial changes) in any implementation version macros? My understanding is that any changes that affect API semantics (including observable side effects) require at least the minor version update.

doc/versioning.rst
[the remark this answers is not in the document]
vmalia: Updated

doc/versioning.rst
vmalia: The existing text from the spec says:
> Increment the:
> - MAJOR version when adding major new functionality and making incompatible API changes, including removing APIs.
> - MINOR version when adding minor functionality and API changes that are backwards compatible.
> - REVISION when **making backwards compatible bug fixes** or any editing change in the document, including minor changes such as correcting typos. Initial REVISION is 1.

My understanding is that the existing spec allows incrementing revision number for bug fixes, while major and minor updates may contain new functionality. This RFC does not modify this behavior.

If we want to remove the revision from macros, then the revision field should explicitly define only documentation changes, without the **bold text** above. Since this will require a separate RFC, I consider this out-of-scope for this proposal.

doc/versioning.rst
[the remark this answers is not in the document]
vmalia: @rscohn2 can you please help understand if my changes are in the right place?

doc/versioning.rst
akukanov: This document describes versioning of the specification, therefore in the context of it "bug fixes" mean changes in the specification, and not in any of the implementations. Therefore, I do not think it says about any new functionality - since bugs fixed in the specification obviously do not involve any source code changes.

As described above, if the specification change includes a new minor functionality or compatible API changes, then the minor version needs to be updated. Therefore I remain positive that new revisions only contain editorial changes but no changes to the API semantics.

I am happy to be corrected, ideally with practical examples. Tagging @rscohn2.

doc/versioning.rst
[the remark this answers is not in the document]
akukanov: I would suggest to go to each component library and propose a patch that describes the spec versioning macro for that very library, This will need to be done anyway I believe, even with a common page like you suggest.

The specification source files that are not specific to any particular library reside directly in the `source` folder.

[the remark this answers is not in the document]
vmalia: Are you suggesting that we add the macro name and its description to each component of the oneAPI spec, like oneMKL and oneTBB? What benefit will we get compared to having it in the common versioning section? My understanding is that we would end up duplicating the content, and any future change would need to happen in multiple places.

doc/versioning.rst
vmalia: I am okay to remove the revision part from the proposal if that was the intended use case for the revision field and is currently being followed. I still think the bold text can be improved to have less ambiguity.

doc/versioning.rst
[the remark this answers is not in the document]
akukanov: To avoid any ambiguity in the macro names and values, just giving a couple of examples is insufficient. In other words, each library specific macro should be well specified. I believe it is best done in the respective sections, to have all relevant information about a library in a single place. It does not preclude describing some general rules in the "shared" part of the specification and referring to that part where needed.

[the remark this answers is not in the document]
vmalia: > ...each library specific macro should be well specified.

I am not sure I understand what you mean by this. Can you provide some example?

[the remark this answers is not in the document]
akukanov: https://eel.is/c++draft/cpp.predefined
https://www.openmp.org/spec-html/5.1/openmpse10.html
https://www.mpi-forum.org/docs/mpi-4.1/mpi41-report/node244.htm

In all cases, the names of the macros are explicitly written, and the values - or rules for building the values - are unambiguous and clear. Well, for OpenMP it is unknown which combination of a year and a month corresponds to which version; I have not looked carefully if this mapping exists somewhere.

Here comes SYCL: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_preprocessor_directives_and_macros

> The version of SYCL defined in this document will have SYCL_LANGUAGE_VERSION substitute the integer 2020, composed with the general SYCL version followed by 2 digits representing the revision number;

This one I do not clearly understand. Is the value 2020 or 202009 or what? But at least, the name is clear.

The rules in this proposal are somewhat ambiguous for both the name and the value, in my view. Yes, one can take an educated guess - but it will still be a guess.

If a clear definition similar to `__cplusplus` and `MPI_VERSION` needs to be given for each oneAPI specification component, in my opinion the best place for it would be the chapter of that component.

doc/versioning.rst
rscohn2: In 4 years, I have only used revision for formatting/editing. I can remove the "compatible buf fixes" clause.

doc/versioning.rst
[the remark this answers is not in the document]
rscohn2: Yes, the original page probably should have been part of the spec, not part of the documnetation. I don't have any strong opinions about whether this should be common or per element. There would not be a lot of duplication or confusion to have it in the element instead of common so I suggest going with Alexey's desire to move it in MKL.

doc/versioning.rst
vmalia: Thank you @rscohn2 . In that case, I will update the RFC to include only major and minor versions in the macro.

doc/versioning.rst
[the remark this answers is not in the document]
vmalia: @akukanov I updated the oneMKL source. Is this better?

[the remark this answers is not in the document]
akukanov: I think it is certainly better than the previous "global" proposal. To me it looks reasonable, but it's now the question for the oneMKL team here. I can think of something similar added for oneDPL and oneTBB.

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
```diff
@@ -11,10 +11,33 @@ This section covers all other features in the design of oneMKL architecture.
...
+Version of oneMKL Specification
 +++++++++++++++++++++++++++++++++++++++++++++
 
-This is the oneMKL specification which is part of the oneAPI specification version 1.0.0.
+This is the oneMKL specification which is part of the oneAPI specification.
+
+The implementation must define a macro, `ONEMKL_SPEC_VERSION` as having the numerical value of the specification version.
+
+The specification version can be created by appending all digits of the specification version in the format of <MAJOR><MINOR>. MINOR version always uses two digits. This version can be used to check the compatibility of the implementation with the specification version.
+
```
spencerpatty, line 22: I have a concern that over time we may reach a MAJOR version larger than 9, what will happen then?  should we always use 4 digits ?  0101,  0102, etc ?

spencerpatty: actually since we are using a macro number through `#define`, I suppose it really doesn't matter, does it ...

spencerpatty: why are we ignoring the revision ?

Rbiessy (approved):

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
vmalia: For the major version, yes, we will keep enabling additional digits as we move beyond `9` or `99`, because it won't affect the numerical comparison. Having a `0` prefix for single digit major versions will need us to update the spec if we ever approach `100`. I did not see a strong reason to do this, so I kept it simple. It is still an option if we feel it will add some advantage.

We are ignoring the revision because as per https://github.com/uxlfoundation/oneAPI-spec/pull/561#discussion_r1725767506, the revision is used only to update the documentation, not the implementation.

vmalia:
@akukanov @mkrainiuk @rscohn2 @spencerpatty if everything looks okay, then please provide your approval. Thank you!

spencerpatty (approved):
Acceptable from my perspective.  Approved

mkrainiuk (approved):

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
```diff
@@ -11,10 +11,40 @@ This section covers all other features in the design of oneMKL architecture.
...
+  | oneAPI 1.2 rev 2 will be represented as a numerical value of 102
+
+Macro Example
+
+.. code-block:: c
+
+  // For oneAPI 1.2 rev 1
+  #define ONEMKL_BLAS_SPEC_VERSION 102
+
+  // For oneAPI 1.2 rev 2
```
mkrainiuk, line 43: Should we mention that `rev 2` is assuming to be the latest one since it's not present in the version macro?

vmalia:
@aelizaro @paveldyakov @lhuot @zettai-reido @sknepper can you please provide your reviews? Thank you!

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
vmalia: Added a line

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
```diff
@@ -11,10 +11,40 @@ This section covers all other features in the design of oneMKL architecture.
...
 +++++++++++++++++++++++++++++++++++++++++++++
 
-This is the oneMKL specification which is part of the oneAPI specification version 1.0.0.
+Each oneMKL domain must define a preprocessor macro to represent the version of the specification that the implementation is compliant with.
+
+The macros for each domain are listed as follows:
+
+  | ONEMKL_BLAS_SPEC_VERSION
+  | ONEMKL_LAPACK_SPEC_VERSION
+  | ONEMKL_SPARSE_SPEC_VERSION
```
spencerpatty, line 23: can we use "SPBLAS" instead of "SPARSE" as someday there will also be SPSOLVERS  sparse solvers and the impls may not be in perfect lock step with SPBLAS ...  but will attempt to be interoperable

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
```diff
@@ -11,10 +11,40 @@ This section covers all other features in the design of oneMKL architecture.
...
-Current Version of this oneMKL Specification
+Specification Version and Compliance
 +++++++++++++++++++++++++++++++++++++++++++++
 
-This is the oneMKL specification which is part of the oneAPI specification version 1.0.0.
+Each oneMKL domain must define a preprocessor macro to represent the version of the specification that the implementation is compliant with.
+
+The macros for each domain are listed as follows:
+
+  | ONEMKL_BLAS_SPEC_VERSION
```
spencerpatty, line 21: If/When we switch to oneMATH name for the Specification and Interfaces project, will we also switch the macro names ?  maybe we want to delay this merge until those are made to not have to deprecate/switch ...  ?

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
vmalia: @spencerpatty To use SPBLAS, we need to define SPBLAS as a domain in the spec's architecture section first, then make the corresponding change here. For now I cannot justify the divergence of the macro name from the domains defined in the spec here: https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.3-rev-1/elements/onemkl/source/architecture/architecture#onemkl-namespaces

If an implementation is not compliant with the spec, then any macro name can be used by that implementation. If your suggestion is to make that change, I can create a new issue, but it would be out of scope for this RFC.

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
vmalia: @spencerpatty The changes you mentioned are from https://github.com/oneapi-src/oneMKL/pull/564 which is an RFC in one of the implementations of this spec. The implementation can choose to take any direction, whether or not it is in compliance with the specification.

@Rbiessy you approved this PR earlier so there are 2 questions - 

1. Does your RFC(https://github.com/oneapi-src/oneMKL/pull/564) block this RFC?
2. Does your RFC aim to make changes to the oneAPI specification eventually? If yes, can you please create a placeholder Issue/PR in this project and mention it here? If not, please note that your oneMKL Interfaces RFC breaks compliance with the oneAPI specification as of the current status of your RFC.

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
spencerpatty: so SPBLAS is a domain and it uses the "sparse" namespace.  Some day in the future there will also be a SPSOLVER  or sparse_solver domain introduced and it will also use the "sparse" namespace ...  I don't see the problem unless you are equating namespace with domain

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
vmalia: I had an offline discussion with the team and we came to an agreement that the `oneMATH` proposal has a lot of work ahead of it, and that name change will have to be made in not just oneMKL Interfaces, but also in multiple places across oneAPI spec and other documents. Since this change will have a wider impact than just this specific part of document where this PR targets, it is better to make that change at that time. It will not affect the core version schema or approach proposed here.
This PR and RFC is NOT blocked by @Rbiessy 's RFC in oneMKL Interfaces.

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
vmalia: @spencerpatty When we add multiple sparse-related functionality, I think it makes sense to make the namespaces more clear, and if required, distinct for each Sparse group of APIs/algorithms, like SPBLAS, SPSOLVER etc.
I would like to refer to the point 5 of the https://github.com/mkrainiuk/oneMKL/blob/rfcs/README.md#rfc-ratification-process which requires any contributor to propose changes to oneAPI spec first, then make those changes in your backend implementation, if you want to comply with the oneAPI specification.

If you think having SPBLAS is better, then please feel free to open a new GH issue for the same to be reflected in all places in the oneAPI specification. Changing existing terminologies for domains and namespaces is out-of-scope for this RFC.

spencerpatty: namespaces won't be changing, but the name of the domains are different and they would share a namespace ... The name of the domain is not SPARSE but SPBLAS,  in some future time the domain of SPSOLVERS would also be added, but both are separate domains that will share a namespace.  I'm not sure why we need to have a GH issue ...  

https://github.com/uxlfoundation/oneAPI-spec/tree/main/source/elements/oneMKL/source/domains/spblas  the folder name of domain is "spblas" and the namespace is "sparse".  The name of the domain is Sparse BLAS here: 
 https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneMKL/source/domains/spblas/spblas.rst of which "SPBLAS" is short.
 
 On the other side, there is this page about "sparse linear algebra" https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneMKL/source/domains/sparse_linear_algebra.inc.rst which points to spblas.rst as the current only domain in sparse linear algebra, but again this is similar to the "dense linear algebra" page containing BLAS and LAPACK domains in it ...  and at some point sparse solvers will be part of the sparse linear algebra set as well.  
 
 All that being said, to me "SPBLAS" should be considered as the default here, and "SPARSE" would be in my mind deviating from existing documentation and specification and require github issue and PR changes...

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
```diff
@@ -11,10 +11,40 @@ This section covers all other features in the design of oneMKL architecture.
...
+  | ONEMKL_LAPACK_SPEC_VERSION
+  | ONEMKL_SPARSE_SPEC_VERSION
+  | ONEMKL_DFT_SPEC_VERSION
+  | ONEMKL_RNG_SPEC_VERSION
```
paveldyakov, line 25: I can see that Stats domain is missed: https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.3-rev-1/elements/onemkl/source/domains/stats/onemkl_stats_overview

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
Rbiessy: It does not matter to me which RFC is implemented first. The RFC for renaming is planning to rename the occurrences of all the macros using `ONEMKL`.
The RFC describes how it impacts the specification. I read the ratification process, it does not say that an issue needs to be created. @mkrainiuk could this be clarified in the RFC process? I created https://github.com/uxlfoundation/oneAPI-spec/issues/580 but I have some concerns the discussions will be split in 2 different places now.

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
vmalia: Good catch. I could not find a corresponding namespace.
https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.3-rev-1/elements/onemkl/source/architecture/architecture
Is this expected?

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
vmalia: I updated the RFC to reflect SPBLAS as the macro name, and we now have some more work to clarify what we want to do with the sparse functionality in the spec itself: https://github.com/uxlfoundation/oneAPI-spec/issues/581

This hopefully unblocks this proposal from that confusion.

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
vmalia: We should discuss this in https://github.com/uxlfoundation/oneAPI-spec/issues/581 as well.

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
mkrainiuk: Hi @Rbiessy, good point, we need to clarify it in the ratification process, and I think any changes related to API should be discussed in Spec only, since opensource is just an implementation of the Spec, I believe it should just follow what was decided for Spec. In this particular case I guess the problem is that we have opensource project migration (which is not related to the Spec) combined with renaming (which is Spec related and should be started in the Spec).

vmalia: Thank you for clarifying, @mkrainiuk . @spencerpatty @Rbiessy Let's proceed with this RFC with the `oneMKL` naming. When the `oneMATH` proposal is made to the oneAPI spec, it can address these macros and other changes across the oneAPI spec simultaneously.

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
```diff
@@ -11,10 +11,41 @@ This section covers all other features in the design of oneMKL architecture.
 
 .. _onemkl_spec_current_version:
 
-Current Version of this oneMKL Specification
+Specification Version and Compliance
 +++++++++++++++++++++++++++++++++++++++++++++
 
-This is the oneMKL specification which is part of the oneAPI specification version 1.0.0.
+Each oneMKL domain must define a preprocessor macro to represent the version of the specification that the implementation is compliant with.
```
al42and, line 17: Do all headers provide this info? E.g.,

- Will `#include "oneapi/mkl.hpp"` provide `ONEMKL_*_SPEC_VERSION` for all domains?
  - If a domain is not supported/installed, will it be undefined of `0`?
- Will `#include "oneapi/mkl/rng.hpp"`  provide `ONEMKL_RNG_SPEC_VERSION`?
- Will `#include "oneapi/mkl/rng.hpp"` provide `ONEMKL_BLAS_SPEC_VERSION`?
- Will `oneapi/mkl/rng/device.hpp` provide `ONEMKL_RNG_SPEC_VERSION`?

vmalia: @al42and good question. The specification(and this RFC) does not define any implementation-specific header files for oneMKL, so the location of the macro can be decided by the implementation itself.

> If a domain is not supported/installed, will it be undefined or 0?

For an undefined domain implementation: https://oneapi-src.github.io/oneMKL/create_new_backend.html#create-wrappers shows how the wrappers in oneMKL Interfaces work and how the implementation can handle errors for unimplemented backends. Note that this is different than the oneAPI specification that we have here. More clarification on that: https://github.com/oneapi-src/oneMKL/blob/develop/README.md#faqs

I added a rule for the macro numbering for the non-compliance of an implementation. Thanks!

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
paveldyakov: Suggestion to add it as "STATS" - such namespace is used for statistics routines

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
```diff
@@ -11,10 +11,41 @@ This section covers all other features in the design of oneMKL architecture.
...
+
+  | oneAPI 1.1 rev 1 will be represented as a numerical value of 101
+  | oneAPI 1.2 rev 1 will be represented as a numerical value of 102
+  | oneAPI 1.2 rev 2 will be represented as a numerical value of 102
+
+Macro Example
+
+.. code-block:: c
+
+  // For oneAPI 1.2 rev 1
```
paveldyakov, line 41: Question: Can domain be aligned with provisional spec version? 
E.g. the latest available release is 1.4. provisional https://github.com/uxlfoundation/oneAPI-spec/releases 

If yes, how we should update the macro in this case?

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
```diff
@@ -11,10 +11,41 @@ This section covers all other features in the design of oneMKL architecture.
...
+
+  | ONEMKL_BLAS_SPEC_VERSION
+  | ONEMKL_LAPACK_SPEC_VERSION
+  | ONEMKL_SPBLAS_SPEC_VERSION
+  | ONEMKL_DFT_SPEC_VERSION
+  | ONEMKL_RNG_SPEC_VERSION
+  | ONEMKL_STAT_SPEC_VERSION
+  | ONEMKL_VM_SPEC_VERSION
+
+The specification version can be created by appending all digits of the specification version in the format of <MAJOR><MINOR>. MINOR version always uses two digits. This version can be used to check the compatibility of the implementation with the specification version. Note that the revision is not included here because it reflects changes only for the specification document without affecting the implementation. If the implementation is not compliant with any release of the specification, then the macro must have a numerical value of `000`.
```
sknepper, line 29: Should we link to a document that describes how/when the MAJOR, MINOR, and revision versions get updated?

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
sknepper, line 17: The oneMKL spec allows for an implementation of the spec to implement just 1 API and throw unsupported for all of the other APIs in that domain. (See https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneMKL/source/architecture/api_design.inc.rst)
In that case, is the implementation considered conformant to whatever version of the spec has the APIs that are included in the implementation (albeit mostly unsupported), or is that implementation considered non-compliant?

vmalia: My understanding is that the implementation will still be compliant, because it follows the definitions of the specification to handle errors for unimplemented APIs.

However, the implementation would not be conformant to the standards that the industry would expect backend libraries to be, because of lack of most APIs.
Alternatively, it could be a feature that a specific hardware-software use case only requires or supports certain APIs as per their app's needs.

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
vmalia: done

paveldyakov: Thanks!

paveldyakov (approved):

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
vmalia: @paveldyakov the provisional versioning (https://github.com/uxlfoundation/oneAPI-spec/blob/main/doc/versioning.rst#provisional-versions) is for specifications that are pending approval.

It is highly unlikely that we will have implementations that will comply with provisional specifications unless the approval of the spec happens later than the implementation in the backend library. If the backend is ready before the spec, the macro can point to a future non-provisional revision because the spec release follows a predictable convention.

From the above link:
> A specification may be published before approval, but must be labeled provisional. Provisional specifications are published as a series of revisions until approval. After approval, the provisional label is removed and the rev is set to 1.

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
vmalia: Added at the end of this section.

akukanov:
For your information, below is what I propose for defining a spec version macro for oneDPL. Use as you like; in case of feedback please comment in the PR #591.
```
A oneDPL implementation must define a preprocessor macro representing the version
of the oneDPL specification that the implementation is compliant with.

.. code:: cpp
  // Defined in <oneapi/dpl/version>
  #define ONEDPL_SPEC_VERSION /*implementation-defined*/

The ``ONEDPL_SPEC_VERSION`` macro must be defined to the decimal literal which value equals to
*major-spec-version* * 100 + *minor-spec-version*, where *major-spec-version* and *minor-spec-version*
are the major and the minor versions of the latest fully supported specification.
[*Example*: ``#define ONEDPL_SPEC_VERSION 104`` for the oneDPL specification 1.4.  -- *end example*]
```

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
al42and: > For an undefined domain implementation, this, oneapi-src.github.io/oneMKL/create_new_backend.html#create-wrappers shows how the wrappers in oneMKL Interfaces work and how the implementation can handle errors for unimplemented backends.

I was referring not to the case of unimplemented _backends_, but to the case of a _domain_ missing entirely: "It is permitted for a library implementation of the oneMKL specification to implement either all, one or more than one of the domains in oneMKL."

But now that I read it, the wording "Each oneMKL domain must define a preprocessor macro [...]" is quite clear that if domain is not supported, the macro is not defined.

vmalia: @al42and yes, you are right.

vmalia:
@rscohn2 we have all the required feedback. Can you please help to merge this PR?

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
paveldyakov: sounds good, thanks!

source/elements/oneMKL/source/architecture/other_architecture.inc.rst
vmalia: The name transition and corresponding changes from this proposal are part of https://github.com/uxlfoundation/oneAPI-spec/pull/596 now.

PR #564, opened 2024-08-15
[oneDPL] Sort By Key APIs

dmitriy-sobolev:
Proposing `oneapi::dpl::sort_by_key` and `oneapi::dpl::stable_sort_by_key` specification for oneDPL.

### DIFF
--- a/source/elements/oneDPL/source/parallel_api.rst
+++ b/source/elements/oneDPL/source/parallel_api.rst
@@ -697,7 +697,7 @@ The elements e of [start, end) must be partitioned with respect to the comparato
         InputValueIt values_first,
         Comparator comp = std::less<typename std::iterator_traits<InputIt>::value_type>{});
 
-``oneapi::dpl::sort_by_key`` sorts a sequence of keys in the range ``[keys_first, keys_last)``,
+``oneapi::dpl::sort_by_key`` sorts a sequence of keys in the range ``[keys_first, keys_last)``
 and simultaneously permutes associated values at the same positions in the range
 ``[values_first, values_first + std::distance(keys_first, keys_last))``
 to match the order of the sorted keys.
@@ -705,8 +705,9 @@ to match the order of the sorted keys.
 Sorting is unstable. That means, keys that do not precede one another according to the given comparator
 and their associated values might be ordered arbitrarily relative to each other.
 
-`comp`` is a function object that satisfies the requirements defined by the `C++ Standard`_
-for the equivalent parameter of `std::sort`.
+``comp`` is a function object that satisfies the requirements defined by the `C++ Standard`_
+for the equivalent parameter of ``std::sort``.
+If no ``comp`` object is provided, the keys are sorted with respect to ``std::less``.
 
 .. code:: cpp
 
@@ -717,7 +718,7 @@ for the equivalent parameter of `std::sort`.
         InputValueIt values_first,
         Comparator comp = std::less<typename std::iterator_traits<InputIt>::value_type>());
 
-``oneapi::dpl::stable_sort_by_key`` sorts a sequence of keys in the range ``[keys_first, keys_last)``,
+``oneapi::dpl::stable_sort_by_key`` sorts a sequence of keys in the range ``[keys_first, keys_last)``
 and simultaneously permutes associated values at the same positions in the range
 ``[values_first, values_first + std::distance(keys_first, keys_last))``
 to match the order of the sorted keys.
@@ -725,8 +726,9 @@ to match the order of the sorted keys.
 Sorting is stable. That means, keys that do not precede one another according to the given comparator
 and their associated values maintain their relative order as in the original sequences.
 
-`comp`` is a function object that satisfies the requirements defined by the `C++ Standard`_
-for the equivalent parameter of `std::sort`.
+``comp`` is a function object that satisfies the requirements defined by the `C++ Standard`_
+for the equivalent parameter of ``std::sort``.
+If no ``comp`` object is provided, the keys are sorted with respect to ``std::less``.
 
 .. _`C++ Standard`: https://isocpp.org/std/the-standard
 .. _`SYCL`: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html
--- a/source/elements/oneDPL/source/parallel_api.rst
+++ b/source/elements/oneDPL/source/parallel_api.rst
@@ -698,11 +698,10 @@ The elements e of [start, end) must be partitioned with respect to the comparato
                 InputValueIt values_first, Comparator comp =
                     std::less<typename std::iterator_traits<InputKeyIt>::value_type>());
 
-``oneapi::dpl::sort_by_key`` sorts a sequence of keys within the range ``[keys_first, keys_last)``.
-It also permutes the corresponding sequence of values within the range
-``[values_first, values_first + distance(keys_first, keys_last))``.
-Each key is associated with a corresponding value,
-and this association is maintained throughout the sorting process.
+``oneapi::dpl::sort_by_key`` sorts a sequence of keys in the range ``[keys_first, keys_last)``,
+and simultaneously permutes associated values at the same positions in the range
+``[values_first, values_first + std::distance(keys_first, keys_last))``
+to match the order of the sorted keys.
 
 The order of equal keys is not guaranteed to be preserved.
 
@@ -720,11 +719,10 @@ If no ``comp`` object is provided, the keys are sorted with respect to ``std::le
                 InputValueIt values_first, Comparator comp =
                     std::less<typename std::iterator_traits<InputKeyIt>::value_type>());
 
-``oneapi::dpl::stable_sort_by_key`` sorts a sequence of keys within the range ``[keys_first, keys_last)``.
-It also permutes the corresponding sequence of values within the range
-``[values_first, values_first + distance(keys_first, keys_last))``.
-Each key is associated with a corresponding value,
-and this association is maintained throughout the sorting process.
+``oneapi::dpl::stable_sort_by_key`` sorts a sequence of keys in the range ``[keys_first, keys_last)``,
+and simultaneously permutes associated values at the same positions in the range
+``[values_first, values_first + std::distance(keys_first, keys_last))``
+to match the order of the sorted keys.
 
 The order of equal keys is preserved.
 
--- a/source/elements/oneDPL/source/parallel_api.rst
+++ b/source/elements/oneDPL/source/parallel_api.rst
@@ -690,11 +690,11 @@ The elements e of [start, end) must be partitioned with respect to the comparato
 
 .. code:: cpp
 
-    template<typename Policy, typename InputKeyIt, typename InputValueIt,
+    template<typename Policy, typename KeyIt, typename ValueIt,
         typename Comparator = std::less<typename std::iterator_traits<InputIt>::value_type>>
     void
-    sort_by_key(Policy&& policy, InputKeyIt keys_first, InputKeyIt keys_last,
-        InputValueIt values_first,
+    sort_by_key(Policy&& policy, KeyIt keys_first, KeyIt keys_last,
+        ValueIt values_first,
         Comparator comp = std::less<typename std::iterator_traits<InputIt>::value_type>());
 
 ``oneapi::dpl::sort_by_key`` sorts the sequence of keys ``[keys_first, keys_last)``
@@ -710,17 +710,17 @@ If no ``comp`` object is provided, keys are sorted with respect to ``std::less``
 Sorting is unstable. That means, keys, which do not precede one another with respect to the given 
 comparator, and their associated values might be ordered arbitrarily relative to each other.
 
-``InputKeyIt`` and ``InputValueIt`` must satisfy the requirements of ``ValueSwappable``,
+``KeyIt`` and ``ValueIt`` must satisfy the requirements of ``ValueSwappable``,
 and ``Comparator`` must satisfy the requirements for the ``Compare`` parameter of ``std::sort``,
 as defined by the `C++ Standard`_.
 
 .. code:: cpp
 
-    template<typename Policy, typename InputKeyIt, typename InputValueIt,
+    template<typename Policy, typename KeyIt, typename ValueIt,
         typename Comparator = std::less<typename std::iterator_traits<InputIt>::value_type>>
     void
-    stable_sort_by_key(Policy&& policy, InputKeyIt keys_first, InputKeyIt keys_last,
-        InputValueIt values_first,
+    stable_sort_by_key(Policy&& policy, KeyIt keys_first, KeyIt keys_last,
+        ValueIt values_first,
         Comparator comp = std::less<typename std::iterator_traits<InputIt>::value_type>());
 
 ``oneapi::dpl::stable_sort_by_key`` sorts the sequence of keys ``[keys_first, keys_last)``
@@ -736,7 +736,7 @@ If no ``comp`` object is provided, keys are sorted with respect to ``std::less``
 Sorting is stable. That means, keys, which do not precede one another with respect to the given 
 comparator, and their associated values maintain the relative order as in the original sequences.
 
-``InputKeyIt`` and ``InputValueIt`` must satisfy the requirements of ``ValueSwappable``,
+``KeyIt`` and ``ValueIt`` must satisfy the requirements of ``ValueSwappable``,
 and ``Comparator`` must satisfy the requirements for the ``Compare`` parameter of ``std::sort``,
 as defined by the `C++ Standard`_.
 
--- a/source/elements/oneDPL/source/parallel_api.rst
+++ b/source/elements/oneDPL/source/parallel_api.rst
@@ -698,12 +698,12 @@ The elements e of [start, end) must be partitioned with respect to the comparato
                 InputValueIt values_first, Comparator comp =
                     std::less<typename std::iterator_traits<InputKeyIt>::value_type>());
 
-:code:`oneapi::dpl::sort_by_key` sorts a sequence of keys in the range :code:`[keys_first, keys_last)`,
+``oneapi::dpl::sort_by_key`` sorts a sequence of keys in the range ``[keys_first, keys_last)``,
 and permutes a sequence of values in the range
-:code:`[values_first, values_first + distance(keys_first, keys_last))` to match the order of the sorted keys.
+``[values_first, values_first + distance(keys_first, keys_last))`` to match the order of the sorted keys.
 The order of equal keys is not guaranteed to be preserved.
 
-If no comparator is provided, the keys are sorted with respect to :code:`std::less{}`.
+If no comparator is provided, the keys are sorted with respect to ``std::less{}``.
 
 .. code:: cpp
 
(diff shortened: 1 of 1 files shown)


source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,39 @@ than an element in the range being searched.
...
+    template<typename Policy, typename InputKeyIt, typename InputValueIt,
+        typename Comparator =
+            std::less<typename std::iterator_traits<InputIt>::value_type>>
+    void
+    sort_by_key(Policy&& policy, InputKeyIt keys_first, InputKeyIt keys_last,
+                InputValueIt values_first, Comparator comp =
+                    std::less<typename std::iterator_traits<InputKeyIt>::value_type>());
+
+``oneapi::dpl::sort_by_key`` sorts a sequence of keys in the range ``[keys_first, keys_last)``,
+and permutes a sequence of values in the range
```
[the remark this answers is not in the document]
andreyfe1: Agree. Maybe it's better to see an example that shows the effect

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,39 @@ than an element in the range being searched.
...
+``[values_first, values_first + distance(keys_first, keys_last))`` to match the order of the sorted keys.
+The order of equal keys is not guaranteed to be preserved.
+
+If no comparator is provided, the keys are sorted with respect to ``std::less{}``.
```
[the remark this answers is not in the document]
andreyfe1: I'd also like to see what `Comparator` is. If it correlates to [C++ Compare](https://en.cppreference.com/w/cpp/named_req/Compare) let's say that

source/elements/oneDPL/source/parallel_api.rst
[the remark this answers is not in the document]
dmitriy-sobolev: Yes, we need it. I initially hesitated since this page does not list such requirements for other algorithms.

source/elements/oneDPL/source/parallel_api.rst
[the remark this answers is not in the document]
dmitriy-sobolev: Thank you both the suggestion. I will clarify the description. 
However, I am not sure about if having an example would be good. It improves understanding, but it is not a common quest of a specification format in my opinion. Let me think about it.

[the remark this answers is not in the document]
dmitriy-sobolev: I've clarified the association between the keys and values. I hope it is sufficient without an example.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,39 @@ than an element in the range being searched.
...
 
+.. code:: cpp
+
+    template<typename Policy, typename InputKeyIt, typename InputValueIt,
+        typename Comparator =
+            std::less<typename std::iterator_traits<InputIt>::value_type>>
+    void
+    sort_by_key(Policy&& policy, InputKeyIt keys_first, InputKeyIt keys_last,
+                InputValueIt values_first, Comparator comp =
+                    std::less<typename std::iterator_traits<InputKeyIt>::value_type>());
```
akukanov, line 699: In the algorithm description below, `std::less{}` is said to be the default comparator. However, since C++14 it means `std::less<void>{}`, which is inconsistent with how the signature is specified. I suggest it to be `std::less` without explicit template argument everywhere. If necessary, the implementation may need to be adjusted accordingly.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,45 @@ than an element in the range being searched.
...
+
+``oneapi::dpl::sort_by_key`` sorts a sequence of keys within the range ``[keys_first, keys_last)``.
+It also permutes the corresponding sequence of values within the range
+``[values_first, values_first + distance(keys_first, keys_last))``.
+Each key is associated with a corresponding value,
+and this association is maintained throughout the sorting process.
```
akukanov, line 705:
```suggestion
``oneapi::dpl::sort_by_key`` sorts a sequence of keys in the range ``[keys_first, keys_last)``,
and simultaneously permutes associated values in the same positions in the range
``[values_first, values_first + std::distance(keys_first, keys_last))`` to match the order of the sorted keys.
```

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,45 @@ than an element in the range being searched.
...
+                InputValueIt values_first, Comparator comp =
+                    std::less<typename std::iterator_traits<InputKeyIt>::value_type>());
+
+``oneapi::dpl::stable_sort_by_key`` sorts a sequence of keys within the range ``[keys_first, keys_last)``.
+It also permutes the corresponding sequence of values within the range
+``[values_first, values_first + distance(keys_first, keys_last))``.
+Each key is associated with a corresponding value,
+and this association is maintained throughout the sorting process.
+
+The order of equal keys is preserved.
```
akukanov, line 729:
```suggestion
Sorting is stable. That means, keys that do not precede one another according to the given comparator
and their associated values maintain their relative order as in the original sequences.
```

source/elements/oneDPL/source/parallel_api.rst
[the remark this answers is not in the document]
dmitriy-sobolev: I've added the description. I haven't provided the signature of the comparator due to its flexibility. I hope providing a reference to the `Compare` named requirement is sufficient.

source/elements/oneDPL/source/parallel_api.rst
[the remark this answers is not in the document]
akukanov: These are the formal requirements for `std::sort` in C++17 (except for those on the comparator which are described separately):

> Requires: RandomAccessIterator shall satisfy the requirements of ValueSwappable. The type of *first shall satisfy the requirements of MoveConstructible and of MoveAssignable.

As far as I can tell, we do not describe that level of detail for other algorithms, and I am not sure we should do it here - except perhaps for `ValueSwappable`, because it does make sense to specify whether the implementations should use `swap` or on `iter_swap`.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,49 @@ than an element in the range being searched.
...
+``oneapi::dpl::sort_by_key`` sorts a sequence of keys within the range ``[keys_first, keys_last)``.
+It also permutes the corresponding sequence of values within the range
+``[values_first, values_first + distance(keys_first, keys_last))``.
+Each key is associated with a corresponding value,
+and this association is maintained throughout the sorting process.
+
+The order of equal keys is not guaranteed to be preserved.
+
+``comp`` is a function object,
+which satisfies the ``Compare`` named requirement defined by the `C++ Standard`_.
```
akukanov, line 710:
```suggestion
``comp`` is a function object that satisfies the requirements defined by the `C++ Standard`_
for the equivalent parameter of `std::sort`.
```
That's because it is arguable if `Compare` is really defined as a named requirement in the standard. Also, explicitly referring to `std::sort` might be useful.

source/elements/oneDPL/source/parallel_api.rst
dmitriy-sobolev: What do you think about this small enhancement?

> Sorting is stable. That means, if keys do not precede one another according to the given comparator, their relative order in the original sequence is preserved, along with their associated values.

It should be more readable due to having "blocks" separated by commas.

dmitriy-sobolev: I've applied the suggestion as is. The sentence is long, but conveys the meaning well.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,39 @@ than an element in the range being searched.
...
+            std::less<typename std::iterator_traits<InputIt>::value_type>>
+    void
+    sort_by_key(Policy&& policy, InputKeyIt keys_first, InputKeyIt keys_last,
+                InputValueIt values_first, Comparator comp =
+                    std::less<typename std::iterator_traits<InputKeyIt>::value_type>());
+
+``oneapi::dpl::sort_by_key`` sorts a sequence of keys in the range ``[keys_first, keys_last)``,
+and permutes a sequence of values in the range
+``[values_first, values_first + distance(keys_first, keys_last))`` to match the order of the sorted keys.
+The order of equal keys is not guaranteed to be preserved.
```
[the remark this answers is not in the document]
dmitriy-sobolev: Applied.

source/elements/oneDPL/source/parallel_api.rst
dmitriy-sobolev: Applied for both `sort_by_key` and `stable_sort_by_key`.

source/elements/oneDPL/source/parallel_api.rst
dmitriy-sobolev: Applied for both algorithms.

source/elements/oneDPL/source/parallel_api.rst
akukanov: The wording modification you suggested seems fine to me as well.

source/elements/oneDPL/source/parallel_api.rst
dmitriy-sobolev: Let's keep it consistent for now, and then change everything in a separate PR.

source/elements/oneDPL/source/parallel_api.rst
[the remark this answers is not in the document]
dmitriy-sobolev: I've added `ValueSwappable`. Note, it was specified for both keys and values to allow swapping values the same way.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,53 @@ than an element in the range being searched.
...
+
+Sorting is stable. That means, keys that do not precede one another according to the given comparator
+and their associated values maintain their relative order as in the original sequences.
+
+``InputKeyIt`` and ``InputValueIt`` must meet the requirements of
+``ValueSwappable`` defined by the `C++ Standard`_.
+
+``comp`` is a function object that satisfies the requirements defined by the `C++ Standard`_
+for the ``Compare`` parameter of ``std::sort``.
+If no ``comp`` object is provided, the keys are sorted with respect to ``std::less``.
```
akukanov, line 737:
```suggestion
``oneapi::dpl::stable_sort_by_key`` sorts the range of keys ``[keys_first, keys_last)``
and simultaneously permutes associated values at the same positions in the range
``[values_first, values_first + std::distance(keys_first, keys_last))``
to match the order of the sorted keys.

Keys are sorted with respect to the provided comparator object ``comp``. That means, for any
two iterators ``i`` and ``j`` to the sorted sequence of keys such that ``i`` precedes ``j``,
``comp(*j, *i) == false``.
If no ``comp`` object is provided, keys are sorted with respect to ``std::less``.

Sorting is stable. That means, keys, which do not precede one another with respect to the given 
comparator, and their associated values maintain the relative order as in the original sequences.

``InputKeyIt`` and ``InputValueIt`` must satisfy the requirements of ``ValueSwappable``,
and ``Comparator`` must satisfy the requirements for the ``Compare`` parameter of ``std::sort``,
as defined by the `C++ Standard`_.
```

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,53 @@ than an element in the range being searched.
 
 The elements e of [start, end) must be partitioned with respect to the comparator used.
 
+.. code:: cpp
+
+    template<typename Policy, typename InputKeyIt, typename InputValueIt,
```
akukanov, line 693: Since, unlike all other described algorithms, there is no output iterators among parameters, perhaps it makes sense to drop "Input" from the parameter names - i.e., to rename as `KeyIt` and `ValueIt`.

dmitriy-sobolev: Thanks, done.

source/elements/oneDPL/source/parallel_api.rst
dmitriy-sobolev: There were more modifications of this part, so this discussion is not relevant anymore.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,57 @@ than an element in the range being searched.
...
+        typename Comparator = std::less<typename std::iterator_traits<InputIt>::value_type>>
+    void
+    sort_by_key(Policy&& policy, KeyIt keys_first, KeyIt keys_last,
+        ValueIt values_first,
+        Comparator comp = std::less<typename std::iterator_traits<InputIt>::value_type>());
+
+``oneapi::dpl::sort_by_key`` sorts the sequence of keys ``[keys_first, keys_last)``
+and simultaneously permutes associated values at the same positions in the range
+``[values_first, values_first + std::distance(keys_first, keys_last))``
+to match the order of the sorted keys.
```
dmitriy-sobolev, line 703: @akukanov, what do you think about adding "in the range" to be more explicit?

```suggestion
``oneapi::dpl::sort_by_key`` sorts the sequence of keys in the range ``[keys_first, keys_last)``
and simultaneously permutes associated values at the same positions in the range
``[values_first, values_first + std::distance(keys_first, keys_last))``
to match the order of the sorted keys.
```

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,57 @@ than an element in the range being searched.
...
+        typename Comparator = std::less<typename std::iterator_traits<InputIt>::value_type>>
+    void
+    stable_sort_by_key(Policy&& policy, KeyIt keys_first, KeyIt keys_last,
+        ValueIt values_first,
+        Comparator comp = std::less<typename std::iterator_traits<InputIt>::value_type>());
+
+``oneapi::dpl::stable_sort_by_key`` sorts the sequence of keys ``[keys_first, keys_last)``
+and simultaneously permutes associated values at the same positions in the range
+``[values_first, values_first + std::distance(keys_first, keys_last))``
+to match the order of the sorted keys.
```
dmitriy-sobolev, line 729:
```suggestion
``oneapi::dpl::stable_sort_by_key`` sorts the sequence of keys in the range ``[keys_first, keys_last)``
and simultaneously permutes associated values at the same positions in the range
``[values_first, values_first + std::distance(keys_first, keys_last))``
to match the order of the sorted keys.
```

source/elements/oneDPL/source/parallel_api.rst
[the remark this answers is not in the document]
dmitriy-sobolev: The note became less obvious after several iterations of changes of the description.

@akukanov, would you mind against extending the brief summary with "That is, the keys and their corresponding values maintain their original association after sorting.", so the note looks like that:

```
``oneapi::dpl::sort_by_key`` sorts the sequence of keys ``[keys_first, keys_last)``
and simultaneously permutes associated values at the same positions in the range
``[values_first, values_first + std::distance(keys_first, keys_last))``
to match the order of the sorted keys.
That is, the keys and their corresponding values maintain their original association after sorting.
```

source/elements/oneDPL/source/parallel_api.rst
akukanov: Actually, I removed something similar because of a confusion for "a/the sequence in the range". I think of ``[keys_first, keys_last)`` as **the** sequence. In other words, "the sequence" and "the range" are mostly interchangeable for me.

source/elements/oneDPL/source/parallel_api.rst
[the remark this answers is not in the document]
akukanov: I would not object to adding this extra sentence if others think it is an improvement.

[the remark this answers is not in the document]
danhoeflinger: I think it is an improvement.  However, maybe a slightly different phrasing would be even more clear, something like:

"That is, a key and its associated value will have the same index in their respective sequences after sorting."

I think this sentence explicitly says the important information that your option implies via the "association".

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,57 @@ than an element in the range being searched.
...
 The elements e of [start, end) must be partitioned with respect to the comparator used.
 
+.. code:: cpp
+
+    template<typename Policy, typename KeyIt, typename ValueIt,
+        typename Comparator = std::less<typename std::iterator_traits<InputIt>::value_type>>
+    void
+    sort_by_key(Policy&& policy, KeyIt keys_first, KeyIt keys_last,
+        ValueIt values_first,
+        Comparator comp = std::less<typename std::iterator_traits<InputIt>::value_type>());
```
danhoeflinger, line 698:
```suggestion
    template<typename Policy, typename KeyIt, typename ValueIt,
        typename Comparator = std::less<typename std::iterator_traits<KeyIt>::value_type>>
    void
    sort_by_key(Policy&& policy, KeyIt keys_first, KeyIt keys_last,
        ValueIt values_first,
        Comparator comp = std::less<typename std::iterator_traits<KeyIt>::value_type>());
```

Similar change required for the signature below.  
I would prefer `{}` or `Comparator{}` as the default parameter to remove confusion, but I think this was already discussed, and intended to change together with the other algorithms.

source/elements/oneDPL/source/parallel_api.rst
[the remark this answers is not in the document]
dmitriy-sobolev: Thanks, I've added this sentence for both sort algorithms.

source/elements/oneDPL/source/parallel_api.rst
dmitriy-sobolev: Applied for both sort algorithms.

dmitriy-sobolev: > I would prefer {} or Comparator{} as the default parameter to remove confusion, but I think this was already discussed, and intended to change together with the other algorithms.

Yes, it was discussed with an agreement to use `()` for the alignment with other algorithms on this page.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,59 @@ than an element in the range being searched.
...
+to match the order of the sorted keys. That is, a key and its associated value
+will have the same index in their respective sequences after sorting.
+
+Keys are sorted with respect to the provided comparator object ``comp``. That means, for any
+two iterators ``i`` and ``j`` to the sorted sequence of keys such that ``i`` precedes ``j``,
+``comp(*j, *i) == false``.
+If no ``comp`` object is provided, keys are sorted with respect to ``std::less``.
+
+Sorting is unstable. That means, keys, which do not precede one another with respect to the given
+comparator, and their associated values might be ordered arbitrarily relative to each other.
```
danhoeflinger, line 712:
```suggestion
Sorting is unstable. In other words, keys which do not precede one another with respect to the given
comparator and their associated values might be ordered arbitrarily relative to each other, which may
not match their original relative ordering.
```

A wording tweak, unnecessary comma removal, and extra clarification for clarity.  

The comma removals  are necessary, the others are optional, you can take if you wish.

source/elements/oneDPL/source/parallel_api.rst
```diff
@@ -688,5 +688,59 @@ than an element in the range being searched.
...
+``[values_first, values_first + std::distance(keys_first, keys_last))``
+to match the order of the sorted keys. That is, a key and its associated value
+will have the same index in their respective sequences after sorting.
+
+Keys are sorted with respect to the provided comparator object ``comp``. That means, for any
+two iterators ``i`` and ``j`` to the sorted sequence of keys such that ``i`` precedes ``j``,
+``comp(*j, *i) == false``.
+If no ``comp`` object is provided, keys are sorted with respect to ``std::less``.
+
+Sorting is stable. That means, keys, which do not precede one another with respect to the given
```
danhoeflinger, line 738: Whatever change is made above should also be made here.

source/elements/oneDPL/source/parallel_api.rst
dmitriy-sobolev: I've removed the commas

source/elements/oneDPL/source/parallel_api.rst
dmitriy-sobolev: I've removed the commas

danhoeflinger (approved):
LGTM, 
best to get @akukanov approval as well prior to merging.

akukanov (approved):

dmitriy-sobolev:
I've merged the main into this branch, because the automatic merge might have resulted in an additional line (probably because of `transform_if` changes earlier at the end of this page).

danhoeflinger (approved):

PR #568, opened 2024-08-21
[oneMKL][RNG] Add integer 64-bit support for uniform distribution

andreyfe1:
Uniform distribution can support `std::int64_t`, `std::uint64_t` also.


iMartyan:
lgtm, thanks for adding

aelizaro (approved):

PR #569, opened 2024-08-23
[oneDPL] Add parallel range algorithms

akukanov:
The specification for oneDPL parallel range algorithms.
oneDPL RFC discussion: https://github.com/oneapi-src/oneDPL/discussions/1741

Tagging @rarutyun @MikeDvorskiy


source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -0,0 +1,321 @@
...
+
+oneDPL provides variations of algorithms that work with ranges defined in the `C++ Standard`_, 6th edition (C++20).
+These algorithms execute according to a oneDPL execution policy supplied as the first argument, similarly to other
+oneDPL algorithms.
+
+The oneDPL parallel range algorithms rely on the functionality of C++20 and are not available in the code
+compiled for earlier editions of the standard.
+
+The parallel range algorithms reside in ``namespace oneapi::dpl::ranges``. Same as the range algorithm functions
+defined by the `C++ standard` in ``namespace std::ranges``, they cannot be found by argument-dependent name lookup.
```
rarutyun, line 16: Do we want just require (on the specification level) that all of them are global callable objects (independently of where C++ standard goes)? With that requirement we give extra benefits for the users.

akukanov: Which specifically benefits would that give? Not theoretical but real ones, something that you can imagine being useful in practice.

rarutyun: passing algorithms as arguments to a function. This is the only practical difference I can see. It's guaranteed to work. With having "special function that..." it's not guaranteed to work. That's it

akukanov: Yes, that's a theoretical possibility. But I do not know why one would use it in practice, given that calling signatures of those algorithms are quite different and no clear reason why someone would want anything like that.
In simple words, I see practical value of this being (nearly) absent.

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -0,0 +1,321 @@
...
+- Both input and output ranges must support random access.
+- For a given algorithm, at least one of the input ranges as well as the output range must be bounded.
+- ``for_each`` does not return its function object.
+
+Whole Sequence Operations
++++++++++++++++++++++++++
+
+.. code:: cpp
+
+  // Defined in <oneapi/dpl/ranges>
```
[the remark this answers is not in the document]
akukanov: Done

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
akukanov: I have added a note about function object based implementations.

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -0,0 +1,323 @@
...
+and cannot be called with explicitly specified template arguments. [*Note*: A typical implementation uses
+predefined function objects which static function call operators have the required signatures. -- *end note*]
+
+The following differences to the standard C++ range algorithms apply:
+
+- The execution policy parameter is added.
+- Output data sequences are defined as ranges, not iterators.
+- Both input and output ranges must support random access.
+- For a given algorithm, at least one of the input ranges as well as the output range must be bounded.
+- ``for_each`` does not return its function object.
```
danhoeflinger, line 26: Do we want to make mention that we won't be supporting `constexpr`, or is that obvious by the intent of the library?

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -0,0 +1,323 @@
...
+              std::indirectly_unary_invocable< std::projected<std::ranges::iterator_t<R>, Proj> > Fn>
+      requires oneapi::dpl::is_execution_policy_v<std::remove_cvref_t<ExecutionPolicy>> &&
+               std::ranges::sized_range<R>
+      std::ranges::borrowed_iterator_t<R>
+        for_each (ExecutionPolicy&& pol, R&& r, Fn f, Proj proj = {});
+
+    // count
+    template <typename ExecutionPolicy, std::ranges::random_access_range R,
+              typename Proj = std::identity,
+              typename T = std::projected_value_t<std::ranges::iterator_t<R>, Proj>>
```
danhoeflinger, line 73: It looks like we are matching the C++26 version of this API with the template ordering and by including this default (though with the mention that the algorithms can't use explicitly specified  template arguments the ordering shouldn't matter).

For my own understanding, when can this default argument for `T` be used?  Won't this type always be deduced from the supplied function argument?


Also, for this PR, should we describe that in at least some cases, we are using APIs from versions beyond C++20?

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -0,0 +1,323 @@
...
+             typename Proj1 = std::identity, typename Proj2 = std::identity>
+      requires oneapi::dpl::is_execution_policy_v<std::remove_cvref_t<ExecutionPolicy>> &&
+               std::ranges::sized_range<R1> && std::ranges::sized_range<R2> &&
+               std::indirectly_comparable< std::ranges::iterator_t<R1>, std::ranges::iterator_t<R2>,
+                                           Pred, Proj1, Proj2 >
+      std::ranges::borrowed_subrange_t<R1>
+        search (ExecutionPolicy&& pol, R1&& r1, R2&& r2, Pred pred = {},
+                Proj1 proj1 = {}, Proj2 proj2 = {});
+
+    // search_n
```
danhoeflinger, line 197: Is there a reason we are including `search_n`, but not all of the `*_n` algorithms?  (no `for_each_n` or copy_n)

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
akukanov: > For my own understanding, when can this default argument for T be used?

It is used when a braced init list is passed as an argument and so type deduction is not possible. As far as I remember, C++26 adds it to all standard algorithms, and also there is a GH issue to consider doing it for oneDPL algorithms.

> Also, for this PR, should we describe that in at least some cases, we are using APIs from versions beyond C++20?

Yes, perhaps the introduction needs to be slightly rephrased in that regard.

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
akukanov: This update matches to what we are adding to the implementation.

Also note that `search_n` is very different semantically from other `*_n` algorithms.

danhoeflinger: Aha, yes I was mistaken about `copy_n` and `for_each_n` because it is listed in the std::ranges namespace, but you are correct that it is only the iterator "flavor" of the API.  I misread the description of `search_n`, but now am reminded of its semantics.

Thanks for the response.

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
danhoeflinger: That makes sense. Thanks.  

I'll leave this unresolved for now to represent the comment about versions of the ranges API from beyond C++20.

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
akukanov: Added a note for this.

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
akukanov: Added a sentence about the signatures matching the current working draft for C++26.

danhoeflinger (approved):
LGTM

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -0,0 +1,327 @@
...
+compiled for earlier editions of the C++ standard.
+
+The parallel range algorithms reside in ``namespace oneapi::dpl::ranges``. Same as the range algorithm functions
+defined by the C++ standard in ``namespace std::ranges``, they cannot be found by argument-dependent name lookup
+and cannot be called with explicitly specified template arguments. [*Note*: A typical implementation uses
+predefined function objects which static function call operators have the required signatures. -- *end note*]
+
+The following differences to the standard C++ range algorithms apply:
+
+- Parallel range algorithms cannot be used in constant expressions.
```
danhoeflinger, line 22: Do we want any mention here that arguments to parallel range algorithms must be sycl device copyable?

There is currently a discussion happening about pointer-to-member fields and functions and their device copyability.  It is currently not specified in the SYCL specification, and functionally has problems when the ABI is not the same between host and device.  With this being a somewhat common pattern in callables arguments to ranges APIs, we may want to mention something here at least about device copyability in general, if not about pointer-to-member fields specifically.

akukanov: Some general wording does make sense, but it is not really specific to parallel range algorithms, so better done in a separate patch.

The pointer-to-member issue seems more of an implementation problem (at least until the SYCL specification says otherwise), so making it a restriction here does not sound right.

danhoeflinger: Yes, I agree that a general comment for oneDPL as a whole about device copyability with device policies is probably the correct approach.

PR #570, opened 2024-08-24
[oneDPL] Clarify the semantics of dpl::begin/end return type

akukanov:
The patch updates the description for buffer wrappers (`dpl::begin/end` function) to clarify the semantics of the type they return.
It also addresses https://github.com/uxlfoundation/oneAPI-spec/issues/482

### DIFF
--- a/source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
+++ b/source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
@@ -43,8 +43,8 @@ Buffer Wrappers
 
 ``oneapi::dpl::begin`` and ``oneapi::dpl::end`` are helper functions
 for passing `SYCL`_ buffers to oneDPL algorithms.
-These functions accept a SYCL buffer and return an object of an unspecified type
-(referred below as *buffer position object*) that satisfies the following requirements:
+These functions accept a SYCL buffer and return a *buffer position object*,
+which type satisfies the following requirements but is otherwise unspecified:
 
 - It is copy-constructible and copy-assignable.
 - It is comparable with ``operator==`` and ``operator!=``.
@@ -86,9 +86,9 @@ for ``begin`` and ``end``. Otherwise, the behavior is undefined.
    auto policy = dpl::execution::dpcpp_default;
 
    auto buf_begin = dpl::begin(buf, sycl::write_only);
-   auto buf_end_1 = dpl::end(buf, sycl::write_only); // matches begin()
+   auto buf_end_1 = dpl::end(buf, sycl::write_only); // arguments match begin()
    dpl::fill(policy, buf_begin, buf_end_1, 42);      // OK
-   auto buf_end_2 = dpl::end(buf, sycl::write_only, sycl::no_init); // does not match begin()
+   auto buf_end_2 = dpl::end(buf, sycl::write_only, sycl::no_init); // arguments do not match
    dpl::fill(policy, buf_begin, buf_end_2, 42);                     // undefined behavior
 
 .. _`SYCL`: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html
--- a/source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
+++ b/source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
@@ -48,7 +48,8 @@ which type satisfies the following requirements but is otherwise unspecified:
 
 - It is copy-constructible and copy-assignable.
 - It is comparable with ``operator==`` and ``operator!=``.
-- It provides the ``get_buffer()`` method that returns the SYCL buffer, which an object was built over.
+- It provides the ``get_buffer()`` method that for a buffer position object returns the SYCL buffer,
+  which the object was built over.
 - The expressions ``a + n`` and ``a - n``, where ``a`` is a buffer position object and ``n``
   is an integer value, are valid and evaluate to a buffer position object such that
   the corresponding position in the buffer follows (precedes) that of ``a`` by ``n``.
--- a/source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
+++ b/source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
@@ -43,33 +43,50 @@ Buffer Wrappers
 
 ``oneapi::dpl::begin`` and ``oneapi::dpl::end`` are helper functions
 for passing `SYCL`_ buffers to oneDPL algorithms.
-These functions accept a buffer and return an object
-of an unspecified type that satisfies the following requirements:
+These functions accept a SYCL buffer and return an object of an unspecified type
+(referred below as *buffer position*) that satisfies the following requirements:
 
-- it is ``CopyConstructible``, ``CopyAssignable``, and comparable
-  with operators ``==`` and ``!=``;
-- the following expressions are valid: ``a + n``, ``a - n``,
-  ``a - b``, where ``a`` and ``b`` are objects of the type,
-  and ``n`` is an integer value;
-- it provides the ``get_buffer()`` method that returns the buffer
-  passed to the ``begin`` or ``end`` function.
+- It is ``CopyConstructible`` and ``CopyAssignable``.
+- It is comparable with operators ``==`` and ``!=``.
+- It provides the ``get_buffer()`` method that returns the SYCL buffer, over which an object was built.
+- The expressions ``a + n`` and ``a - n``, where ``a`` is a buffer position object and ``n``
+  is an integer value, are valid and evaluate to a buffer position object representing
+  a position in the buffer that follows or precedes the position of ``a`` by ``n``.
+  If this new position is out of the buffer bounds, the behavior is undefined.
+- The expression ``a - b``, where ``a`` and ``b`` are buffer position objects,
+  is valid and evaluates to an integer value ``n`` such that ``a == b + n``.
+
+These operations are only valid for objects built over the same SYCL buffer.
+If an expression operates with buffer position objects that are not built
+over the same buffer, the behavior is undefined.
 
 When invoking an algorithm, the buffer passed to ``begin`` should be the same
 as the buffer passed to ``end``. Otherwise, the behavior is undefined.
 
+An algorithm may return a buffer position object, which can then be compared and used
+in expressions with other objects built over the same buffer.
+
+.. code:: cpp
+
+   namespace dpl = oneapi::dpl; // see "Namespaces"
+   sycl::buffer buf {/*...*/};
+   auto pos = dpl::find(dpl::execution::dpcpp_default, dpl::begin(buf), dpl::end(buf), value);
+   int value_index = (pos == dpl::end(buf))? -1 : pos - dpl::begin(buf);
+
 SYCL deduction tags (the ``TagT`` parameters) and ``sycl::property::no_init`` 
-allow to specify an access mode to be used by algorithms for accessing the buffer.
+allow to specify an access mode to be used by algorithms for accessing a SYCL buffer.
 The mode serves as a hint, and can be overridden depending on semantics of the algorithm.
 When invoking an algorithm, the same access mode arguments should be used
 for ``begin`` and ``end``. Otherwise, the behavior is undefined.
 
 .. code:: cpp
-      
-      using namespace oneapi;
-      auto buf_begin = dpl::begin(buf, sycl::write_only);
-      auto buf_end_1 = dpl::end(buf, sycl::write_only);
-      auto buf_end_2 = dpl::end(buf, sycl::write_only, sycl::no_init);
-      dpl::fill(dpl::execution::dpcpp_default, buf_begin, buf_end_1, 42); // allowed
-      dpl::fill(dpl::execution::dpcpp_default, buf_begin, buf_end_2, 42); // not allowed
+
+   namespace dpl = oneapi::dpl;
+   sycl::buffer buf {/*...*/};
+   auto buf_begin = dpl::begin(buf, sycl::write_only);
+   auto buf_end_1 = dpl::end(buf, sycl::write_only);
+   auto buf_end_2 = dpl::end(buf, sycl::write_only, sycl::no_init);
+   dpl::fill(dpl::execution::dpcpp_default, buf_begin, buf_end_1, 42); // allowed
+   dpl::fill(dpl::execution::dpcpp_default, buf_begin, buf_end_2, 42); // not allowed
 
 .. _`SYCL`: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html
--- a/source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
+++ b/source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
@@ -44,11 +44,11 @@ Buffer Wrappers
 ``oneapi::dpl::begin`` and ``oneapi::dpl::end`` are helper functions
 for passing `SYCL`_ buffers to oneDPL algorithms.
 These functions accept a SYCL buffer and return an object of an unspecified type
-(referred below as *buffer position*) that satisfies the following requirements:
+(referred below as *buffer position object*) that satisfies the following requirements:
 
 - It is ``CopyConstructible`` and ``CopyAssignable``.
-- It is comparable with operators ``==`` and ``!=``.
-- It provides the ``get_buffer()`` method that returns the SYCL buffer, over which an object was built.
+- It is comparable with ``operator==`` and ``operator!=``.
+- It provides the ``get_buffer()`` method that returns the SYCL buffer, which an object was built over.
 - The expressions ``a + n`` and ``a - n``, where ``a`` is a buffer position object and ``n``
   is an integer value, are valid and evaluate to a buffer position object representing
   a position in the buffer that follows or precedes the position of ``a`` by ``n``.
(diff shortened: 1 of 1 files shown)


source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,50 @@ Buffer Wrappers
...
-      auto buf_end_1 = dpl::end(buf, sycl::write_only);
-      auto buf_end_2 = dpl::end(buf, sycl::write_only, sycl::no_init);
-      dpl::fill(dpl::execution::dpcpp_default, buf_begin, buf_end_1, 42); // allowed
-      dpl::fill(dpl::execution::dpcpp_default, buf_begin, buf_end_2, 42); // not allowed
+
+   namespace dpl = oneapi::dpl;
+   sycl::buffer buf {/*...*/};
+   auto buf_begin = dpl::begin(buf, sycl::write_only);
+   auto buf_end_1 = dpl::end(buf, sycl::write_only);
+   auto buf_end_2 = dpl::end(buf, sycl::write_only, sycl::no_init);
```
danhoeflinger, line 88: Optional:
It may be more clear to name these `buf_end` and `buf_end_no_init`, but its just my preference.

danhoeflinger (approved):
LGTM, optional variable name change.

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,50 @@ Buffer Wrappers
...
+
+These operations are only valid for objects built over the same SYCL buffer.
+If an expression operates with buffer position objects that are not built
+over the same buffer, the behavior is undefined.
 
 When invoking an algorithm, the buffer passed to ``begin`` should be the same
 as the buffer passed to ``end``. Otherwise, the behavior is undefined.
 
+An algorithm may return a buffer position object, which can then be compared and used
+in expressions with other objects built over the same buffer.
```
rarutyun, line 67:
```suggestion
in expressions with other objects built over the same SYCL buffer.
```

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,50 @@ Buffer Wrappers
...
 .. code:: cpp
-      
-      using namespace oneapi;
-      auto buf_begin = dpl::begin(buf, sycl::write_only);
-      auto buf_end_1 = dpl::end(buf, sycl::write_only);
-      auto buf_end_2 = dpl::end(buf, sycl::write_only, sycl::no_init);
-      dpl::fill(dpl::execution::dpcpp_default, buf_begin, buf_end_1, 42); // allowed
-      dpl::fill(dpl::execution::dpcpp_default, buf_begin, buf_end_2, 42); // not allowed
+
+   namespace dpl = oneapi::dpl;
```
rarutyun, line 84: line 71: just to not forget.

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,50 @@ Buffer Wrappers
...
 
+An algorithm may return a buffer position object, which can then be compared and used
+in expressions with other objects built over the same buffer.
+
+.. code:: cpp
+
+   namespace dpl = oneapi::dpl; // see "Namespaces"
+   sycl::buffer buf {/*...*/};
+   auto pos = dpl::find(dpl::execution::dpcpp_default, dpl::begin(buf), dpl::end(buf), value);
+   int value_index = (pos == dpl::end(buf))? -1 : pos - dpl::begin(buf);
```
rarutyun, line 74:
```suggestion
   int value_index = (pos == dpl::end(buf)) ? -1 : (pos - dpl::begin(buf));
```

Maybe, it's more readable

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
rarutyun: Agreed, I also like this suggestion more

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,50 @@ Buffer Wrappers
...
-      dpl::fill(dpl::execution::dpcpp_default, buf_begin, buf_end_1, 42); // allowed
-      dpl::fill(dpl::execution::dpcpp_default, buf_begin, buf_end_2, 42); // not allowed
+
+   namespace dpl = oneapi::dpl;
+   sycl::buffer buf {/*...*/};
+   auto buf_begin = dpl::begin(buf, sycl::write_only);
+   auto buf_end_1 = dpl::end(buf, sycl::write_only);
+   auto buf_end_2 = dpl::end(buf, sycl::write_only, sycl::no_init);
+   dpl::fill(dpl::execution::dpcpp_default, buf_begin, buf_end_1, 42); // allowed
+   dpl::fill(dpl::execution::dpcpp_default, buf_begin, buf_end_2, 42); // not allowed
```
rarutyun, line 90: How about:

```suggestion
   dpl::fill(dpl::execution::dpcpp_default, buf_begin, buf_end_2, 42); // not allowed, Undefined Behavior
```
or even just `// Undefined Behavior`, without _not allowed_

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,50 @@ Buffer Wrappers
...
+   sycl::buffer buf {/*...*/};
+   auto pos = dpl::find(dpl::execution::dpcpp_default, dpl::begin(buf), dpl::end(buf), value);
+   int value_index = (pos == dpl::end(buf))? -1 : pos - dpl::begin(buf);
+
 SYCL deduction tags (the ``TagT`` parameters) and ``sycl::property::no_init`` 
-allow to specify an access mode to be used by algorithms for accessing the buffer.
+allow to specify an access mode to be used by algorithms for accessing a SYCL buffer.
 The mode serves as a hint, and can be overridden depending on semantics of the algorithm.
 When invoking an algorithm, the same access mode arguments should be used
 for ``begin`` and ``end``. Otherwise, the behavior is undefined.
```
rarutyun, line 80: I am wondering why the behavior is undefined. Looking in the actual implementation I kind of understand why, but I believe we could do a better job.

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,50 @@ Buffer Wrappers
...
-  with operators ``==`` and ``!=``;
-- the following expressions are valid: ``a + n``, ``a - n``,
-  ``a - b``, where ``a`` and ``b`` are objects of the type,
-  and ``n`` is an integer value;
-- it provides the ``get_buffer()`` method that returns the buffer
-  passed to the ``begin`` or ``end`` function.
+These functions accept a SYCL buffer and return an object of an unspecified type
+(referred below as *buffer position*) that satisfies the following requirements:
+
+- It is ``CopyConstructible`` and ``CopyAssignable``.
```
rarutyun, line 49:
```suggestion
- It satisfies ``CopyConstructible`` and ``CopyAssignable`` C++ named requirements.
```

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,50 @@ Buffer Wrappers
...
+- It is comparable with operators ``==`` and ``!=``.
```
rarutyun, line 50: Not sure if it's better or not but how about this?
```suggestion
- It is comparable with ``operator==`` and ``operator!=``.
```

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,50 @@ Buffer Wrappers
...
+- It provides the ``get_buffer()`` method that returns the SYCL buffer, over which an object was built.
```
rarutyun, line 51:
```suggestion
- It provides the ``get_buffer()`` method that returns the SYCL buffer, which an object was built over.
```

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,50 @@ Buffer Wrappers
...
+- The expressions ``a + n`` and ``a - n``, where ``a`` is a buffer position object and ``n``
+  is an integer value, are valid and evaluate to a buffer position object representing
+  a position in the buffer that follows or precedes the position of ``a`` by ``n``.
```
rarutyun, line 54: Too many "position" words, though I am not sure how to make it better. Maybe:
```suggestion
  is an integer value, are valid and evaluate to a buffer position object representing
  an offset in the buffer that follows or precedes the position of ``a`` by ``n``.
```

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,50 @@ Buffer Wrappers
...
 
 When invoking an algorithm, the buffer passed to ``begin`` should be the same
 as the buffer passed to ``end``. Otherwise, the behavior is undefined.
 
+An algorithm may return a buffer position object, which can then be compared and used
+in expressions with other objects built over the same buffer.
+
+.. code:: cpp
+
+   namespace dpl = oneapi::dpl; // see "Namespaces"
```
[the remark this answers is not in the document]
akukanov: Technically, we do not. But it does not hurt to remind of this in the context of the example code.

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
akukanov: There are two external requirements to deal with:
- Access mode is a template parameter of `sycl::accessor`.
- Standard C++ algorithms (those in `std`) require the same type for their "first" and "last" parameters.

Therefore the "access mode" argument of `dpl::begin` has to become part of its return value type so that accessors can be created. And then the return type of `dpl::end` needs to match that of `begin` to better comply with the C++ standard.

One way to "do a better job" is to use "sentinels" for the end-of-sequence arguments, i.e. do not require types to match. Another one is to deprecate access modes and `no_init` for `dpl::begin/end`. In both cases, it is not just a "quality of implementation" question.

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,50 @@ Buffer Wrappers
...
+- The expressions ``a + n`` and ``a - n``, where ``a`` is a buffer position object and ``n``
+  is an integer value, are valid and evaluate to a buffer position object representing
+  a position in the buffer that follows or precedes the position of ``a`` by ``n``.
+  If this new position is out of the buffer bounds, the behavior is undefined.
+- The expression ``a - b``, where ``a`` and ``b`` are buffer position objects,
+  is valid and evaluates to an integer value ``n`` such that ``a == b + n``.
+
+These operations are only valid for objects built over the same SYCL buffer.
+If an expression operates with buffer position objects that are not built
+over the same buffer, the behavior is undefined.
```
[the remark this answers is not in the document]
akukanov: The first sentence might look excessive, but in fact it explicitly says which expressions are *valid*, therefore clarifying the wording above such as "The expression ``a - b`` ... is valid".

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
akukanov: Changed in a different way.

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
akukanov: Changed in a slightly different way

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
akukanov: I also do not quite like it, and also have no good idea how to improve.

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
akukanov: I added more comments to address the clarity concern.

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
danhoeflinger: location?  index?
also "in the buffer that follows" could be may be better as "into the buffer that follows" if we use "offset" or "index"

akukanov: The reading problem here is really caused by "a buffer position object representing a position in the buffer", which sounds like tautology. But without it, the wording would become sloppy:

> evaluate to a buffer position object that follows or precedes ``a`` by ``n``.

because what is that "follows/precedes" relation between the *objects* (and not positions)? We would need to introduce it first.

 And if we change to "a buffer position object representing an index into the buffer", then why is not it a buffer *index* object? :) Same for "offset", and also offsets do not follow or precede one another. And I would not speak of an element either, because the object gives no direct way to access an element or even to get its index.

danhoeflinger: I think something like "a buffer position object representing an index location into the buffer" would be reasonable and get the point the point across without seeming to define a word with itself.

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,50 @@ Buffer Wrappers
...
 
 When invoking an algorithm, the buffer passed to ``begin`` should be the same
 as the buffer passed to ``end``. Otherwise, the behavior is undefined.
 
+An algorithm may return a buffer position object, which can then be compared and used
+in expressions with other objects built over the same buffer.
+
```
danhoeflinger, line 68: Would it be helpful to explicitly mention that there is **no** requirement that these types are dereferenceable on the host, and there is **no** requirement that a buffer index is retrievable from these objects?

In the documentation, I think it would be useful to call this out, in a specification, I'm not sure.  
That is clearly left unspecified already, but explicit mention may give more clarity about the intentions for how these buffer position objects are to be used (and perhaps more importantly how they are not intended to be used).

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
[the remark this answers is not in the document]
akukanov: I have changed my mind and will use a similar wording :)

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
akukanov: Let me try this:
> ... evaluate to a buffer position object such that the corresponding position in the buffer follows or precedes that of ``a`` by ``n``.

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
akukanov: Added a note to clarify the intent.

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
rarutyun: I believe, adding it explicitly doesn't hurt and doesn't create the tautology case, but fine.

rarutyun (approved):
I think it's good enough

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
```diff
@@ -43,33 +43,56 @@ Buffer Wrappers
...
-  ``a - b``, where ``a`` and ``b`` are objects of the type,
-  and ``n`` is an integer value;
-- it provides the ``get_buffer()`` method that returns the buffer
-  passed to the ``begin`` or ``end`` function.
+These functions accept a SYCL buffer and return a *buffer position object*,
+which type satisfies the following requirements but is otherwise unspecified:
+
+- It is copy-constructible and copy-assignable.
+- It is comparable with ``operator==`` and ``operator!=``.
+- It provides the ``get_buffer()`` method that returns the SYCL buffer, which an object was built over.
```
danhoeflinger, line 51:
```suggestion
- It provides the ``get_buffer()`` method that returns the SYCL buffer, which the object was built over.
```

danhoeflinger (approved):
Other than the small edit suggestion above, this LGTM.

source/elements/oneDPL/source/parallel_api/buffer_wrappers.rst
danhoeflinger: I see your point, we are talking about the type requirements, not requirements on objects of that type.

Do we need to clarify this a little more to say 
"Objects of this type provide the member function ``get_buffer()`` that returns the SYCL buffer over which the object was built."  ?

If so, perhaps the line above it should be
"Objects of this type are comparable with ``operator==`` and ``operator!``."


I guess what I take issue with in the current formulation is that there is not a notion that `get_buffer()` connects directly to the calling object.  It can be inferred, but I don't think it is stated with this language.

akukanov: Should be better now :)

PR #571, opened 2024-08-28
[oneDPL] Add Histogram API

danhoeflinger:
Proposing oneapi::dpl::histogram specification for oneDPL.

Version 1a from discussion.

PR #572, opened 2024-08-28
[oneDPL] Histogram spec version 1b (draft)

danhoeflinger:
Draft for discussion purposes.
This version uses the function signature from oneDPL existing implementation.  Alternate of #546, which contains signature which has been modified to be narrower.

Version 1b from email.


danhoeflinger:
We have decided to go with #571  instead of this.  closing.

PR #575, opened 2024-09-06
[oneDPL] Create dedicated pages for iterators and algorithms

akukanov:
This patch completes restructuring of oneDPL Parallel API part into smaller pages.

The content is unchanged, so the 0-day review period applies (https://github.com/uxlfoundation/uxl_operational_procedures/blob/release/Process_Documents/Organization_Operational_Process.md#review--approval-process).


danhoeflinger (approved):
LGTM, confirmed the sections are unchanged.

PR #583, opened 2024-09-23
[oneMKL][spblas] Add sorted_by_rows property

Rbiessy:
Following the discussion in https://github.com/oneapi-src/oneMKL/pull/527#discussion_r1666587321 I suggest to introduce the `sorted_by_rows` property.

Note that this [section](https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneMKL/source/domains/spblas/data_types/format-descriptions.rst?plain=1#L29) already specifies that the default state of COO format is unsorted.

See compiled html page: [html.zip](https://github.com/user-attachments/files/17101211/html.zip)

### DIFF
--- a/source/elements/oneMKL/source/domains/spblas/data_types/set_matrix_property.rst
+++ b/source/elements/oneMKL/source/domains/spblas/data_types/set_matrix_property.rst
@@ -52,8 +52,8 @@ details the property for each matrix format:
         - Guarantees that the column indices are sorted in ascending order for
           a given row.
       * - COO
-        - Guarantees that the indices are sorted by row then by column in
-          ascending order.
+        - Guarantees that the indices are sorted by rows then by columns within
+          rows in ascending order.
 
 sorted_by_rows
 ^^^^^^^^^^^^^^
@@ -72,7 +72,7 @@ details the property for each matrix format:
           :ref:`oneapi::mkl::invalid_argument<onemkl_exception_invalid_argument>`
           exception.
       * - COO
-        - Guarantees that the indices are sorted by row in ascending order.
+        - Guarantees that the indices are sorted by rows in ascending order.
           The column indices within a row do not need to be sorted.
 
 set_matrix_property
--- a/source/elements/oneMKL/source/domains/spblas/data_types/set_matrix_property.rst
+++ b/source/elements/oneMKL/source/domains/spblas/data_types/set_matrix_property.rst
@@ -29,31 +29,51 @@ Matrix properties
    holds all the given properties. A property can be set as a hint for backends
    to optimize some operations. Multiple properties can be set to the same handle.
 
+symmetric
+^^^^^^^^^
+
+Guarantees that the user-provided matrix data are symmetric, meaning the matrix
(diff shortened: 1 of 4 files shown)


source/elements/oneMKL/source/domains/spblas/data_types/set_matrix_property.rst
```diff
@@ -39,12 +40,20 @@ Matrix properties
...
             column in ascending order.
+      * - ``sorted_by_rows``
+        - | Guarantees that the user-provided matrix data is partially sorted.
+            Has no effect if the property ``sorted`` is set.
+          | For COO this guarantees that the indices are sorted by row in
+            ascending order. The column indices within a row do not need to be
+            sorted.
+          | Setting this property for a matrix using the CSR format will throw an
+            :ref:`oneapi::mkl::invalid_argument<onemkl_exception_invalid_argument>`
+            exception.
```
gajanan-choudhary, line 56: Why should it throw an `invalid_argument` exception for CSR format? This is the default state of a CSR format anyway. Setting this property on a CSR matrix should do nothing regardless of whether the `sorted` property has been set or not.

Rbiessy: I don't really think of the CSR as being sorted by rows in the sense that the values in `row_ptr` don't represent indices and can't be ordered in any other way. I understand the idea for allowing this. I think not throwing an exception is a risk for inexperienced users that maybe wanted to set the `sorted` property instead.

gajanan-choudhary: > I think not throwing an exception is a risk for inexperienced users that maybe wanted to set the sorted property instead.

Good thinking. This is fine then.

Rbiessy:
Update the table to fix https://github.com/uxlfoundation/oneAPI-spec/issues/588

New html doc: [docs.zip](https://github.com/user-attachments/files/17151858/docs.zip)

source/elements/oneMKL/source/domains/spblas/data_types/set_matrix_property.rst
```diff
@@ -28,23 +29,51 @@ Matrix properties
...
+      :widths: 20 80
+
+      * - Matrix format
+        - Description
+      * - CSR
+        - Guarantees that the column indices are sorted in ascending order for
+          a given row.
+      * - COO
+        - Guarantees that the indices are sorted by row then by column in
+          ascending order.
```
gajanan-choudhary, line 56:
```suggestion
        - Guarantees that the indices are sorted by rows then by columns
          within rows in ascending order.
```

gajanan-choudhary (approved):
LGTM

source/elements/oneMKL/source/domains/spblas/data_types/set_matrix_property.rst
Rbiessy: Done in https://github.com/uxlfoundation/oneAPI-spec/pull/583/commits/eef74a30ab15edbfed9e1d470913aec66ec23903

PR #591, opened 2024-09-27
[oneDPL] Add requirements to header files and specification version macro

akukanov:


source/elements/oneDPL/source/common.rst
```diff
@@ -2,14 +2,51 @@
...
+.. code:: cpp
+
+  // Defined in <oneapi/dpl/version>
+
+  #define ONEDPL_SPEC_VERSION /*implementation-defined*/
+
+The ``ONEDPL_SPEC_VERSION`` macro must be defined to the decimal literal which value equals to
+*major-spec-version* * 100 + *minor-spec-version*, where *major-spec-version* and *minor-spec-version*
+are the major and the minor versions of the latest fully supported specification.
+[*Example*: ``#define ONEDPL_SPEC_VERSION 104`` for the oneDPL specification 1.4.  -- *end example*]
```
vmalia, line 52: I don't think there is a oneDPL specification - it is only oneAPI specification, which defines versioning in terms of major, minor and revision, not in the format `1.4`. Can you please clarify?

akukanov: With all that said, if a proposal comes how the wording can be improved, I will happily consider and apply it.

vmalia: @akukanov 
>> By default, elements do not have independent versioning. ... Other elements can be split out and have independent versioning if the need arises with agreement of the oneAPI core team.

> and there were precedents in the past: Level Zero and oneVPL, before being removed from the oneAPI spec, maintained their versions independently.

I would extend the question - if we separate so many chapters, what will remain under the oneAPI umbrella, considering the software architecture: https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.3-rev-1/architecture#software-architecture ?

akukanov: @vmalia is it a "what if" question? Maybe I do not get it, but so far nothing is separated, is it?

vmalia: Currently we have the same version numbering and scheme used across the entire oneAPI spec, except Level Zero like you mentioned. As I see you are suggesting splitting oneDPL's versioning scheme from oneAPI-spec, is that right?

My question is - what is stopping other chapters from doing the same thing? Is there a specific reason/method we use to decide if a chapter is eligible for "independent versioning"?

akukanov: > As I see you are suggesting splitting oneDPL's versioning scheme from oneAPI-spec, is that right?

No, I am not. May I ask which exactly part of the wording led you to such a conclusion?
I just find it fully appropriate to refer to the oneDPL specification and its version, whether it is aligned with the version of the whole oneAPI specification (as now) or not. More generally, I view the oneAPI specification as modular, not monolithic. Each element chapter is the specification for that element/component, while the oneAPI specification is a collection of those.

> what is stopping other chapters from doing the same thing? Is there a specific reason/method we use to decide if a chapter is eligible for "independent versioning"?

As I pointed out, I interpret the versioning rules as allowing, in principle, independent versioning of the specifications for oneAPI elements. The method, I believe, will be just as for any other specification change - someone need to make an RFC proposal stating the reasons as well as some analysis, and get the RFC approved. And I interpret "with agreement of the oneAPI core team" in the quote above as "must be approved by the core team".

source/elements/oneDPL/source/common.rst
```diff
@@ -2,14 +2,51 @@
...
+A oneDPL implementation must define a preprocessor macro representing the version
+of the oneDPL specification that the implementation is compliant with.
+
+.. code:: cpp
+
+  // Defined in <oneapi/dpl/version>
+
+  #define ONEDPL_SPEC_VERSION /*implementation-defined*/
+
+The ``ONEDPL_SPEC_VERSION`` macro must be defined to the decimal literal which value equals to
```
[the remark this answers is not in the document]
akukanov: Done.

danhoeflinger (approved):
LGTM.  
Obviously, we need to resolve the discussion of ONEAPI_SPEC_VERSION vs ONEDPL_SPEC_VERSION,  once that is resolved I think this is good to go.

akukanov:
> Obviously, we need to resolve the discussion of ONEAPI_SPEC_VERSION vs ONEDPL_SPEC_VERSION, once that is resolved I think this is good to go.

@danhoeflinger I think the macro name is not the point of the discussion. That question was covered at https://github.com/uxlfoundation/oneAPI-spec/pull/561#discussion_r1721912875 and resolved to use library-specific prefixes. The  oneMKL team went even further and decided to use domain-specific prefixes - see the code in #561.

So the discussion above is not about naming the macro but about the wording around.

danhoeflinger:
> 
> @danhoeflinger I think the macro name is not the point of the discussion. That question was covered at [#561 (comment)](https://github.com/uxlfoundation/oneAPI-spec/pull/561#discussion_r1721912875) and resolved to use library-specific prefixes. The oneMKL team went even further and decided to use domain-specific prefixes - see the code in #561.
> 
> So the discussion above is not about naming the macro but about the wording around it.

Aha, yes.  Understood.  Thanks for the pointer.

source/elements/oneDPL/source/common.rst
akukanov: Added the following wording to the beginning of the oneDPL chapter:

> This is the specification of oneAPI DPC++ Library, part of the oneAPI Specification version 1.4-provisional-rev-1.

Hopefully this clarifies the intent.

vmalia: >> As I see you are suggesting splitting oneDPL's versioning scheme from oneAPI-spec, is that right?

> No, I am not. May I ask which exactly part of the wording led you to such a conclusion?
```rst
[*Example*: ``#define ONEDPL_SPEC_VERSION 104`` for the oneDPL specification 1.4.  -- *end example*]
```
It was not obvious to me that oneAPI and oneDPL (or any other chapter names) can be used interchangeably when it comes to versioning. It should be clarified somewhere. That being said, I am okay with the idea overall.

vmalia (approved):

timmiesmith (approved):
Looks good to me. Thank you.

Issue #271, opened 2020-11-18
DPL: Constructors and operator= for DPC++ execution policies are not explicitly specified


andreyfe1:
The following are not explicitly specified for DPC++ execution policies:

- copy constructor
- move constructor
- copy assignment
- move assignment

Is their behavior considered to be default?
For example. What if an implementation decide to mark operator= as deleted, will it be considered as a deviation from the Spec?

rarutyun:
From my perspective if we don't explicitly list the default members the behavior is as they are there with `= default` marker. The argument for that is the destructor. I believe that we didn't specify the destructor almost anywhere. But it's valid operation since we can create the objects on the stack (e.g. `device_policy`) and this is the desired scenario.

Even if we don't specify the layout explicitly I think we should consider prohibiting those default members as the breaking change. Consider one more example: Imagine that the class has all default members specified as `= default` without specifying the layout. I think user would assume that they can be called. Then, on the next version of the product the implementation has been changed and contains the reference in the layout which breaks the ability to call copy-assignment. To me it also looks like breaking change.

akukanov:
I am sure marking copy/move operations as `= default` is not a good idea, because it might unnecessarily limit the implementation options, and it really does not mean much because the internals of a device policy are not exposed.

I think the best course of actions is to add the operations to the class and describe the expected semantics.

Tagging @rarutyun @danhoeflinger

rarutyun:
> I think the best course of actions is to add the operations to the class and describe the expected semantics.

I don't feel strongly either way (status quo vs adding the operators with expected semantics). Perhaps, what you are suggesting is better to avoid ambiguity in the behavior. The thing is I saw several examples in C++ standard where "default" members are not specified for the class in the synopsis, so everybody just assumes they have a `= default` behavior (though, it might be hard to know exactly what's going on because, as you pointed out, there is no much information about class internals). I don't exactly remember, whether (for such cases I am talking about) there are some semantical requirements for "default" members.

Again, to avoid ambiguity what you are suggesting sounds reasonable.

Going back to my example with a destructor, I didn't find any mentioning of `= default` for destructor in C++ standard. It's explicitly defined in the synopsis when it has something "special": if it's non-trivial, `virtual`, `protected`, etc.

danhoeflinger:
I suppose the way to think about this might be:
1) Does the specification explicitly require any of these to be defined in a particular way to make sense?
[...]

PR #594, opened 2024-10-03
[oneDPL] Specify copy and move semantics for device_policy

akukanov:
Addresses #271.

For this patch, I believe the 7 days review period [applies](https://github.com/uxlfoundation/uxl_operational_procedures/blob/release/Process_Documents/Organization_Operational_Process.md#review--approval-process):

> 7 days for major changes to non-API modifications (such as changes to the documentation of an API)

### DIFF
--- a/source/elements/oneDPL/source/parallel_api/execution_policies.rst
+++ b/source/elements/oneDPL/source/parallel_api/execution_policies.rst
@@ -41,10 +41,10 @@ with the `C++ Standard`_, 6th edition (C++20):
 
 See "Execution policies" in the `C++ Standard`_ for more information.
 
-Device Execution Policy
-+++++++++++++++++++++++
+Device Execution Policies
++++++++++++++++++++++++++
 
-A device execution policy class ``oneapi::dpl::execution::device_policy`` specifies
+The device execution policy class ``oneapi::dpl::execution::device_policy`` specifies
 the `SYCL`_ device and queue to run oneDPL algorithms.
 
 .. code:: cpp
@@ -75,8 +75,6 @@ the `SYCL`_ device and queue to run oneDPL algorithms.
       }
     }
   }
-
-``dpcpp_default`` is a predefined execution policy object to run algorithms on the default SYCL device.
 
 device_policy Class
 ^^^^^^^^^^^^^^^^^^^
@@ -116,8 +114,7 @@ A policy instance constructed as or assigned a copy of another instance is assoc
 a ``sycl::queue`` that compares equal, as defined by `SYCL`_, to the queue of that other instance.
 A policy instance constructed as or assigned a move of another instance is associated with the queue
 of that other instance without copying it. It is unspecified whether the original policy is associated
-with any queue after the move. [*Note*: The predefined ``dpcpp_default`` policy object is immutable
-(``const``) and cannot be moved. -- *end note*]
+with any queue after the move.
 
 .. code:: cpp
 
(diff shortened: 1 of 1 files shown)


source/elements/oneDPL/source/parallel_api/execution_policies.rst
```diff
@@ -101,15 +102,23 @@ device_policy Class
...
 
 The ``KernelName`` template parameter, also aliased as ``kernel_name`` within the class template,
 is to explicitly provide a name for SYCL kernels executed by an algorithm the policy is passed to.
 
+The ``device_policy`` type is copy constructible, copy assignable, move constructible and move assignable.
+A policy instance constructed as or assigned a copy of another instance is associated with
+a ``sycl::queue`` that compares equal, as defined by `SYCL`_, to the queue of that other instance.
+A policy instance constructed as or assigned a move of another instance is associated with the queue
+of that other instance without copying it. It is unspecified whether the original policy is associated
+with any queue after the move. [*Note*: The predefined ``dpcpp_default`` policy object is immutable
```
andreyfe1, line 119: I think you can write a **Note** as it's done here: https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneDPL/source/index.rst

andreyfe1 (approved):
LGTM. Thanks!

source/elements/oneDPL/source/parallel_api/execution_policies.rst
akukanov: Thanks for reviewing Andrey.

For a short clarification that depends on the context of the preceding wording an "inlined" note (similar to what the C++ standard uses) seems better to me. I will think more of it though.

andreyfe1: Thanks. It looks nicer now

source/elements/oneDPL/source/parallel_api/execution_policies.rst
```diff
@@ -101,15 +99,22 @@ device_policy Class
...
 
 The ``KernelName`` template parameter, also aliased as ``kernel_name`` within the class template,
 is to explicitly provide a name for SYCL kernels executed by an algorithm the policy is passed to.
 
+The ``device_policy`` type is copy constructible, copy assignable, move constructible and move assignable.
+A policy instance constructed as or assigned a copy of another instance is associated with
+a ``sycl::queue`` that compares equal, as defined by `SYCL`_, to the queue of that other instance.
+A policy instance constructed as or assigned a move of another instance is associated with the queue
+of that other instance without copying it. It is unspecified whether the original policy is associated
+with any queue after the move.
```
danhoeflinger, line 116:
```suggestion
of that other instance without copying it. It is unspecified whether the original policy is associated
with any queue after being moved from.
```
I don't really like use "the move" in this context without first introducing the move we are referring to more directly.

danhoeflinger (approved):
LGTM, small wording suggestion, not required for me.

source/elements/oneDPL/source/parallel_api/execution_policies.rst
akukanov: Would this be better?

> It is unspecified whether the moved-from policy instance is associated with any queue after the move.

danhoeflinger: yes that is better

timmiesmith (approved):
This looks good to me.

PR #595, opened 2024-10-09
Extend parallel_reduce named requirements to allow rvalue reduction

kboyarinov:
Extend `PararallelReduceFunc` and `ParallelReduceReduction` to allow user callable objects accepting rvalues.

### DIFF
--- a/source/elements/oneTBB/source/named_requirements/algorithms/par_reduce_func.rst
+++ b/source/elements/oneTBB/source/named_requirements/algorithms/par_reduce_func.rst
@@ -13,12 +13,18 @@ A type `Func` satisfies `ParallelReduceFunc` if it meets the following requireme
 
 **ParallelReduceFunc Requirements: Pseudo-Signature, Semantics**
 
+.. cpp:function:: Value Func::operator()(const Range& range, Value&& x) const
+
+or
+
 .. cpp:function:: Value Func::operator()(const Range& range, const Value& x) const
 
     Accumulates result for a subrange, starting with initial value ``x``.
     ``Range`` type must meet the :doc:`Range requirements <range>`.
     ``Value`` type must be the same as a corresponding template parameter for the
     :doc:`parallel_reduce algorithm <../../algorithms/functions/parallel_reduce_func>` algorithm.
+
+    If both variations are provided, the implementation should prefer the first form.
 
 See also:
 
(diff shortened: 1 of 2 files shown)


source/elements/oneTBB/source/named_requirements/algorithms/par_reduce_func.rst
```diff
@@ -13,13 +13,19 @@ A type `Func` satisfies `ParallelReduceFunc` if it meets the following requireme
...
+or
+
 .. cpp:function:: Value Func::operator()(const Range& range, const Value& x) const
 
     Accumulates result for a subrange, starting with initial value ``x``.
     ``Range`` type must meet the :doc:`Range requirements <range>`.
     ``Value`` type must be the same as a corresponding template parameter for the
     :doc:`parallel_reduce algorithm <../../algorithms/functions/parallel_reduce_func>` algorithm.
 
+    If both variations are provided, the implementation should prefer the first form.
```
akukanov, line 27: This is not fully clear to me. In which sense an implementation should **prefer** one form over the other? Are there certain conditions when each form can be used? What happens if an implementation does not follow this advice?

akukanov: I wonder if this difference is / should be covered by how we [describe](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.3-rev-1/elements/onetbb/source/named_requirements#named-requirements) *pseudo-signatures*:

> `bool operator<(const T &x, const T &y)`
> A real signature may differ from the pseudo-signature that it implements in ways where implicit conversions would deal with the difference. For an example type U, the real signature that implements operator< in the table above can be expressed as `int operator<( U x, U y )`, because C++ permits implicit conversion from int to bool, and implicit conversion from U to (const U&). Similarly, the real signature `bool operator<( U& x, U& y )` is acceptable because C++ permits implicit addition of a const qualifier to a reference type.

Essentially, a pseudo-signature in named requirements describes the intended use of types by the library. Perhaps this should be updated for C++11 so that the difference between rvalues and lvalues is taken into account.

akukanov: https://github.com/uxlfoundation/oneAPI-spec/pull/647 tries to improve the description of pseudo-signatures so that they are more applicable for setting requirements on value categories.

PR #600, opened 2024-11-05
[oneMKL][LAPACK] Fix ldc size requirements in {or,un}mtr related files

sknepper:
For ormtr and unmtr, the specification incorrectly specifies that LDC must be >= max(1,n), when in fact it should be >= max(1,m).
This was done for all {or,un}mtr related APIs (buffer, USM, and scratchpad size).
Also changed notation in these files to ld* >= x instead of x <= ld*.
Fixed an issue where the first row of the input parameters table had been bolded (incorrectly considered a header row) and added missing table to unmtr_scratchpad_size.

### DIFF
--- a/source/elements/oneMKL/source/domains/lapack/ormtr.rst
+++ b/source/elements/oneMKL/source/domains/lapack/ormtr.rst
@@ -54,7 +54,7 @@ In the descriptions below, ``r`` denotes the order of :math:`Q`:
 .. container:: tablenoborder
 
      .. list-table:: 
-        :header-rows: 1
+        :header-rows: 0
 
         * -  :math:`r = m` 
           -  if ``side = side::left`` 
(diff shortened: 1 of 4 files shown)


ericlars:
Did you build it? The only change I don't know the effect of is

:header-rows: 0

sknepper:
> Did you build it? The only change I don't know the effect of is
> 
> :header-rows: 0

Yes, I built it locally. That change says there are no header rows. With the current code, you can see here: https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.3-rev-1/elements/onemkl/source/domains/lapack/ormtr
that the first row of the input parameter table about r is bolded (compare the formatting of "if" in the two lines).
Thanks for the question!

ericlars:
Gotcha, thanks

ericlars (approved):

PR #603, opened 2024-12-10
Update non-compliance version to allow definition checks

vmalia:
## Problem Statement
When using a definition check,
```c
#ifdef ONEMATH_BLAS_SPEC_VERSION
```
a preprocessor may evaluate an undefined macro to `0`. This means that the existing `000` version for non-compliance is equivalent to the macro not being defined at all and yields the same result for `#ifdef`.

## Solution
To enable the check for definition of the macro itself, meaning, provide different outputs for:
```c
#define ONEMATH_BLAS_SPEC_VERSION 0   // domain non-compliant with specification
```
case and
```c
// #define ONEMATH_BLAS_SPEC_VERSION   // missing macro definition, may be 0
```
case, we need to use a number other than `0`. `001` is hence proposed as the non-compliance version. It does not conflict with any existing release of the specification. As a natural extension, versions between `001` and `100`(Spec release 1.0), can be treated as alpha/beta and used as per special cases that can be defined later as and if needed.

Thanks to @zettai-reido for detecting this issue.

### DIFF
--- a/source/elements/oneMath/source/architecture/other_architecture.inc.rst
+++ b/source/elements/oneMath/source/architecture/other_architecture.inc.rst
@@ -26,7 +26,7 @@ The macros for each domain are listed as follows:
   | ONEMATH_STATS_SPEC_VERSION
   | ONEMATH_VM_SPEC_VERSION
 
-The specification version can be created by appending all digits of the specification version in the format of <MAJOR><MINOR>. MINOR version always uses two digits. This version can be used to check the compatibility of the implementation with the specification version. Note that the revision is not included here because it reflects changes only for the specification document without affecting the implementation. If the implementation is not compliant with any release of the specification, then the macro must have a numerical value of `000`.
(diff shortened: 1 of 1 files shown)


vmalia:
@rscohn2 @zettai-reido please approve.

zettai-reido:
> I have no issues with this change but I don't understand the motivation for it.
> 
> ```
> a preprocessor may evaluate an undefined macro to 0
> ```
> 
> Can you elaborate more on this? This sounds like a compiler issue or a different issue. Which compiler are you using and do you have a reproducer?
> 
> This example shows that the `ifdef` branch should only be taken if a macro is defined on clang: https://gcc.godbolt.org/z/6P6rh5cv5

Please try simpler one:
```
#if X > 0
#warning "X > 0"
#else
#warning "X == 0"
#endif
```
https://gcc.gnu.org/onlinedocs/cpp/If.html

Rbiessy:
Thanks for the explanation, that's good to know!
The situation could always be de-ambiguated by checking if `defined(X)` but maybe it is easier if the version is set to a positive integer.

Rbiessy (approved):

vmalia:
@rscohn2 can you please help with the merge?

Rbiessy:
Is this urgent? I am also able to merge.

vmalia:
@Rbiessy please go ahead. Not urgent, but trying to close on the spec versioning, since there are oneMKL product changes pending on this PR.

PR #604, opened 2024-12-11
[oneMath][RNG] Add geometric distribution to Device API

iMartyan:
Proposal to add [geometric distribution](https://en.wikipedia.org/wiki/Geometric_distribution) to oneMath RNG Device API.
Overview: API similar to existing Bernoulli distribution. Supported data types: 32 and 64 integer types.

_fix:_ changed notes style for RNG

### DIFF
--- a/source/elements/oneMath/source/domains/rng/device_api/device-distributions.rst
+++ b/source/elements/oneMath/source/domains/rng/device_api/device-distributions.rst
@@ -60,7 +60,7 @@ basic random number generators.
      -     Description
    * -     :ref:`onemath_device_rng_uniform_discrete`
      -     integer
-     -     float
+     -     floating point
      -     Uniform discrete distribution on the interval [``a,b``)
    * -     :ref:`onemath_device_rng_bits`
      -     integer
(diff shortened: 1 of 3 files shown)


source/elements/oneMath/source/domains/rng/device_api/device-distributions.rst
```diff
@@ -78,6 +78,10 @@ basic random number generators.
      -     integer
      -     integer
      -     Bernoulli distribution
+   * -     :ref:`onemath_device_rng_geometric`
+     -     integer
+     -     float
```
aelizaro, line 83: double-checking: is it `float` for both `int32_t`/`int64_t` or should it be `float`/`double` ?

iMartyan: You are correct, `int64_t` mapped to `double` in terms of type.
Here, as an idea, `float` refers to general floating-point data type, not exact `float` type.
The same is done for discrete uniform distr and `integer` type in general in this table.

Please share if you think it make sense to reformat this table

andreyfe1 (approved):
Looks good to me

source/elements/oneMath/source/domains/rng/device_api/device-distributions.rst
aelizaro: hmmm, in the table with Continuous distributions it states float/double separately. Probably it makes sense to change `float` to `floating point` and add a note that it can vary for distributions implementation depending on the integer data type size

iMartyan: Applied. @aelizaro please take a look one more time :)

PR #614, opened 2025-02-28
[oneDPL] Add more parallel range algorithms

akukanov:
As a follow-up to the oneDPL RFC proposal https://github.com/uxlfoundation/oneDPL/pull/2037, this PR adds new parallel range algorithm signatures to the specification: `fill`, ~`generate`~, `move`, `replace`, `replace_if`, `remove`, `remove_if`, `mismatch`, `minmax_element`, `minmax`, `min`, `max`, `find_first_of`, `find_end`, `is_sorted_until`.


source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -324,5 +440,79 @@ Mutating Operations
...
+      requires oneapi::dpl::is_execution_policy_v<std::remove_cvref_t<ExecutionPolicy>> &&
+               std::ranges::sized_range<R> && std::invocable<F&> &&
+               std::indirectly_writable<std::ranges::iterator_t<R>, std::invoke_result_t<F&>>
+      std::ranges::borrowed_iterator_t<R>
+        generate (ExecutionPolicy&& pol, R&& r, F gen);
+
+    // replace
+    template <typename ExecutionPolicy, std::ranges::random_access_range R,
+              typename Proj = std::identity,
+              typename T1 = std::projected_value_t<std::ranges::iterator_t<R>, Proj>, typename T2 = T1>
```
MikeDvorskiy, line 472: I think a default type written as `std::projected_value_t` should be omitted(here and in the other places), because it available since C++26. 
Also that difference should be added to the differences list above.

akukanov: Cool. The problem is, we already use it for `count`, `find`, and `search_n`.

MikeDvorskiy: For example,
https://github.com/uxlfoundation/oneDPL/pull/2100/files#diff-c7a5edf4a7b057df5179c8f71fb4701c66ea627fb1aaf8965e4c2e81f0c43a94R815

akukanov: The definition of `std::projected_value_t` only uses APIs available in C++20, so I think the proper resolution is to "define" a custom `projected_value_t` in the spec (and in the implementation as well).

akukanov: https://github.com/uxlfoundation/oneAPI-spec/pull/619 is the PR to fix the issue.

danhoeflinger (approved):
LGTM,
I confirmed that these signatures match p3179r7, other than the noted differences involving constexpr, and sizing of both ranges for `mismatch`.

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -344,5 +460,79 @@ Mutating Operations
...
+      requires oneapi::dpl::is_execution_policy_v<std::remove_cvref_t<ExecutionPolicy>> &&
+               std::ranges::sized_range<R> && std::invocable<F&> &&
+               std::indirectly_writable<std::ranges::iterator_t<R>, std::invoke_result_t<F&>>
+      std::ranges::borrowed_iterator_t<R>
+        generate (ExecutionPolicy&& pol, R&& r, F gen);
+
+    // replace
+    template <typename ExecutionPolicy, std::ranges::random_access_range R,
+              typename Proj = std::identity,
+              typename T1 = /*projected-value-type*/<std::ranges::iterator_t<R>, Proj>, typename T2 = T1>
```
danhoeflinger, line 492: For my own understanding...
What is the point of defaulted template arguments here without default function arguments (wont they always be overridden by the deduced type)?

Are there situations where this may resolve some ambiguity when attempting to deduce the types `T1`, `T2` from the function arguments?

Is it merely to describe the "expected type"?

danhoeflinger: Ah, nevermind. I found a couple answers.  One would be in the context of a function pointer, another would be providing arguments with something like this `{}` which can only be deduced with a default template arg.

akukanov:
The most recent change removes `generate` from the list, as there are unresolved concerns for it in the C++ LWG review of P3179.

MikeDvorskiy (approved):
LGTM

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -206,6 +276,32 @@ Sequence Search and Comparison
       bool equal (ExecutionPolicy&& pol, R1&& r1, R2&& r2, Pred pred = {},
                   Proj1 proj1 = {}, Proj2 proj2 = {});
 
+    // mismatch
+    template<typename ExecutionPolicy, std::ranges::random_access_range R1,
+             std::ranges::random_access_range R2, typename Pred = std::ranges::equal_to,
+             typename Proj1 = std::identity, typename Proj2 = std::identity>
+      requires oneapi::dpl::is_execution_policy_v<std::remove_cvref_t<ExecutionPolicy>> &&
+               (std::ranges::sized_range<R1> || std::ranges::sized_range<R2>) &&
```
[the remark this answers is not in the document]
akukanov: We will keep it like that, it's described in the differences - see https://github.com/uxlfoundation/oneAPI-spec/pull/610/files

rarutyun (approved):

danhoeflinger (approved):
Reapproving after removal of generate

PR #615, opened 2025-02-28
update pre-commit and python deps

rscohn2:
Update requirements.txt, following directions from requirements.header.txt

Update pre-commit with:
```
pre-commit update all
```


rscohn2:
closing python venv was contaminated

PR #616, opened 2025-03-05
[oneMath][RNG] Added random engine adaptors

andreyfe1:
New engine adaptor is added. It wraps a random engine to add extra functionality: to query how many random numbers were taken from an engine during the `generate` call.


source/elements/oneMath/source/domains/rng/device_api/device-rng-count_engine_adaptor.rst
```diff
@@ -0,0 +1,94 @@
...
+..
+.. SPDX-License-Identifier: CC-BY-4.0
+
+.. _onemath_device_rng_count_engine_adaptor:
+
+Count Engine Adaptor
+====================
+
+``count_engine_adaptor`` is a random number engine adaptor that counts how many
+random numbers were taken from an engine during the ``generate`` call.
```
aelizaro, line 11: Should we specify what is meant by `random number` here: if it is 32-bit generator output or something else? As different engines may provide different bits output.
Possible suggestions:
* how many times 32 random bits were taken from the engine
* how many times the random engine state is advanced
* introduce public operator() or generate() for engine and count its calls

andreyfe1: I like the option `how many times 32 random bits were taken from the engine` or `how many times the random engine state is advanced`, but still don't see the best option to write. Will think about it

aelizaro (approved):

aelizaro (approved):

PR #618, opened 2025-03-20
[oneMath][RNG] Make number of generators for MT2203 be implementation defined

andreyfe1:
In general, the amount of Mersenne Twister pseudorandom number generators mustn't be specified as concrete number. Implementors need to have freedom in terms of how much Mersenne Twister generators to use


aelizaro (approved):

paveldyakov (commented):
Thank you for the changes, Andrey!

PR #619, opened 2025-03-25
[oneDPL] Fix the function signatures to not require std::projected_value_t

akukanov:
Another oversight found in the signatures of range algorithms: `std::projected_value_t` requires C++26.

To overcome the issue, it is replaced with an analogous alias added only for exposition (i.e., not as a public API) and defined in terms of C++20.

For comparison, this is the standard definition of `std::projected_value_t`:

```c++
template<indirectly_readable I, indirectly_regular_unary_invocable<I> Proj>
using projected_value_t = // freestanding
  remove_cvref_t<invoke_result_t<Proj&, iter_value_t<I>&>>;
```

It's a backward-compatible bug fix. Ideally it should be added to a revision of the specification v1.4.

### DIFF
--- a/source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
+++ b/source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
@@ -47,7 +47,7 @@ of parallel range algorithms.
 
    // C++20 analogue of std::projected_value_t; exposition only
    template<typename I, typename Proj>
-   using projected-value-type = std::remove_cvref_t<std::invoke_result_t<Proj&, std::iter_value_t<I>&>>;
+   using /*projected-value-type*/ = std::remove_cvref_t<std::invoke_result_t<Proj&, std::iter_value_t<I>&>>;
 
 Whole Sequence Operations
 +++++++++++++++++++++++++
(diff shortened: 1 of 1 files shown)


source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -37,6 +37,18 @@ The following differences to the standard C++ range algorithms apply:
...
++++++++++++++++++++++
+
+The following auxiliary entities are only defined for the purpose of exposition, to aid the specification
+of parallel range algorithms.
+
+.. code:: cpp
+
+   // C++20 analogue of std::projected_value_t; exposition only
+   template<typename I, typename Proj>
+   using projected-value-type = std::remove_cvref_t<std::invoke_result_t<Proj&, std::iter_value_t<I>&>>;
```
rarutyun, line 50: I understand that it's perhaps hard to make exposition only stuff italic in RST but how about the following?

```suggestion
   using /* projected-value-type */ = std::remove_cvref_t<std::invoke_result_t<Proj&, std::iter_value_t<I>&>>;
```

akukanov: Applied.

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -37,6 +37,18 @@ The following differences to the standard C++ range algorithms apply:
 Except for these differences, the signatures of parallel range algorithms correspond to the working draft
 of the next edition of the C++ standard (C++26).
 
+Auxiliary definitions
```
danhoeflinger, line 40:
```suggestion
Auxiliary Definitions
```

danhoeflinger (approved):
Other than minor nitpick it LGTM

MikeDvorskiy (approved):
LGTM

PR #620, opened 2025-04-03
[oneDPL] Indirectly Device Accessible Iterator Customization Point and Public Trait

danhoeflinger:
Specification for "indirectly device accessible iterator" customization point and public trait.  Based upon proposed [RFC].(https://github.com/uxlfoundation/oneDPL/pull/1999)

### DIFF
--- a/source/elements/oneDPL/source/parallel_api/execution_policies.rst
+++ b/source/elements/oneDPL/source/parallel_api/execution_policies.rst
@@ -99,11 +99,11 @@ to run algorithms on a SYCL device. When an algorithm runs with ``device_policy`
 it is capable of processing SYCL buffers (passed via ``oneapi::dpl::begin/end``),
 data in the host memory and data in Unified Shared Memory (USM), including device USM.
 Data placed in the host memory and USM can be passed to oneDPL algorithms
-as pointers and random access iterators. oneDPL provides some :ref:`iterators <iterators>` which are
-compatible with algorithms when using a ``device_policy``. Custom iterators may also be used but users should
+as pointers and random access iterators. oneDPL provides some :ref:`iterators <iterators>` that are
+compatible with algorithms when using a ``device_policy``. Custom iterators may also be used, but users should
 ensure that those iterators have a defined :ref:`"passed directly" customization point <iterators-passed-directly>`
-to avoid unnecessary data movement. The iterators must also be SYCL device copyable to be used with
-algorithms using a ``device_policy``. The way to transfer data from the host memory
+to avoid unnecessary data movement. The iterators must also be SYCL device-copyable to be used with
+algorithms utilizing a ``device_policy``. The way to transfer data from the host memory
 to a device and back is unspecified; per-element data movement to/from a temporary storage
 is a possible valid implementation.
 
--- a/source/elements/oneDPL/source/parallel_api/execution_policies.rst
+++ b/source/elements/oneDPL/source/parallel_api/execution_policies.rst
@@ -99,12 +99,8 @@ to run algorithms on a SYCL device. When an algorithm runs with ``device_policy`
 it is capable of processing SYCL buffers (passed via ``oneapi::dpl::begin/end``),
 data in the host memory and data in Unified Shared Memory (USM), including device USM.
 Data placed in the host memory and USM can be passed to oneDPL algorithms
-as pointers and random access iterators. oneDPL provides some :ref:`iterators <iterators>` that are
-compatible with algorithms when using a ``device_policy``. Custom iterators may also be used, but users should
-ensure that those iterators have a defined :ref:`"passed directly" customization point <iterators-passed-directly>`
-to avoid unnecessary data movement. The iterators must also be SYCL device-copyable to be used with
-algorithms utilizing a ``device_policy``. The way to transfer data from the host memory
-to a device and back is unspecified; per-element data movement to/from a temporary storage
+as pointers and random access iterators, please see :ref:`iterators <iterators>`. The way to transfer data from the host
+memory to a device and back is unspecified; per-element data movement to/from a temporary storage
 is a possible valid implementation.
 
 The ``KernelName`` template parameter, also aliased as ``kernel_name`` within the class template,
--- a/source/elements/oneDPL/source/parallel_api/execution_policies.rst
+++ b/source/elements/oneDPL/source/parallel_api/execution_policies.rst
@@ -99,9 +99,9 @@ to run algorithms on a SYCL device. When an algorithm runs with ``device_policy`
 it is capable of processing SYCL buffers (passed via ``oneapi::dpl::begin/end``),
 data in the host memory and data in Unified Shared Memory (USM), including device USM.
 Data placed in the host memory and USM can be passed to oneDPL algorithms
-as pointers and random access iterators, please see :ref:`iterators <iterators>`. The way to transfer data from the host
-memory to a device and back is unspecified; per-element data movement to/from a temporary storage
-is a possible valid implementation.
+as pointers and random access iterators, including :doc:`oneDPL iterators <iterators>`. The way to transfer data from
+the host memory to a device and back is unspecified; per-element data movement to/from a temporary storage is a possible
+valid implementation.
 
 The ``KernelName`` template parameter, also aliased as ``kernel_name`` within the class template,
 is to explicitly provide a name for SYCL kernels executed by an algorithm the policy is passed to.
--- a/source/elements/oneDPL/source/parallel_api/execution_policies.rst
+++ b/source/elements/oneDPL/source/parallel_api/execution_policies.rst
@@ -99,7 +99,11 @@ to run algorithms on a SYCL device. When an algorithm runs with ``device_policy`
 it is capable of processing SYCL buffers (passed via ``oneapi::dpl::begin/end``),
 data in the host memory and data in Unified Shared Memory (USM), including device USM.
 Data placed in the host memory and USM can be passed to oneDPL algorithms
-as pointers and random access iterators. The way to transfer data from the host memory
+as pointers and random access iterators. oneDPL provides some :ref:`iterators <iterators>` which are
+compatible with algorithms when using a ``device_policy``. Custom iterators may also be used but users should
+ensure that those iterators have a defined :ref:`"passed directly" customization point <iterators-passed-directly>`
+to avoid unnecessary data movement. The iterators must also be SYCL device copyable to be used with
+algorithms using a ``device_policy``. The way to transfer data from the host memory
 to a device and back is unspecified; per-element data movement to/from a temporary storage
 is a possible valid implementation.
 
--- a/source/elements/oneDPL/source/parallel_api/iterators.rst
+++ b/source/elements/oneDPL/source/parallel_api/iterators.rst
@@ -362,8 +362,9 @@ using the set of source iterators provided.
 
 Other Iterators
 +++++++++++++++
-Pointers are assumed to be USM shared or device memory pointers and are *indirectly device accessible*.
-Pointers are trivially copyable and therefore SYCL device-copyable.
+Pointers are assumed to be USM shared or device memory pointers when used in combination with an algorithm with a
+``device_policy`` and are *indirectly device accessible*. Pointers are trivially copyable and therefore SYCL
+device-copyable.
 
 ``std::reverse_iterator<IteratorT>`` is an ``AdaptingIteratorSource`` if ``IteratorT`` is an ``AdaptingIteratorSource``.
 ``std::reverse_iterator<IteratorT>`` is an *indirectly device accessible iterator* if ``IteratorT`` is an *indirectly
--- a/source/elements/oneDPL/source/parallel_api/iterators.rst
+++ b/source/elements/oneDPL/source/parallel_api/iterators.rst
@@ -379,7 +379,7 @@ Pointers are trivially copyable and therefore SYCL device-copyable.
 Customization For User Defined Iterators
 ++++++++++++++++++++++++++++++++++++++++
 oneDPL provides a mechanism to indicate whether custom iterators are *indirectly device accessible* by defining an
- Argument-Dependent Lookup (ADL) based customization point, ``is_onedpl_indirectly_device_accessible``.
+Argument-Dependent Lookup (ADL) based customization point, ``is_onedpl_indirectly_device_accessible``.
 
 ADL-based Customization Point: ``is_onedpl_indirectly_device_accessible``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/source/elements/oneDPL/source/parallel_api/iterators.rst
+++ b/source/elements/oneDPL/source/parallel_api/iterators.rst
@@ -11,19 +11,22 @@ Iterators
 Requirements For Iterator Use With Device Policies
 ++++++++++++++++++++++++++++++++++++++++++++++++++
 Iterator are, by default, not assumed to refer to content which is accessible on the device which requires that content
-to be copied to the device prior to being used inside a `SYCL`_ kernel. We define iterators to be "indirectly device
-accessible iterators" when they can inherently be dereferenced on the device in a SYCL kernel. When using oneDPL
-algorithms with a ``device_policy``, "indirectly device accessible iterators" avoid unnecessary data movement when
-provided as input or output arguments.
+to be copied to the device prior to being used inside a `SYCL`_ kernel. We define the term "indirectly device
+accessible" to mean representing data which is accessible on the device within a SYCL kernel. "Indirectly device
+accessible iterators" are iterators which can inherently be dereferenced on the device within a SYCL kernel.
+:doc:`*buffer position objects* <buffer_wrappers>` returned by ``oneapi::dpl::begin`` and ``oneapi::dpl::end`` are not
+iterators, but they are "indirectly device accessible" because they represent data which is accessible on the device.
+When using oneDPL algorithms with a ``device_policy``, "indirectly device accessible" types avoid unnecessary data
+movement when provided as input or output arguments.
 
-Examples of "indirectly device accessible iterators" include SYCL USM shared or device memory, or iterator types like
+Examples of "indirectly device accessible" iterators include SYCL USM shared or device memory, or iterator types like
 ``counting_iterator`` or ``discard_iterator`` that do not require any data to be copied to the device to be used in a
-SYCL kernel. An example of an iterator which does not refer to "indirectly device accessible" is a ``std::vector``
-iterator, which requires the data to be copied to the device in some way prior to usage in a SYCL kernel within
-algorithms used with a ``device_policy``.
+SYCL kernel. An example of an iterator type which is not "indirectly device accessible" is a ``std::vector`` iterator,
+with a host allocator which requires the data to be copied to the device in some way prior to usage in a SYCL kernel
+within algorithms used with a ``device_policy``.
 
-When used as input or output with algorithms using a ``device_policy`` "indirectly device accessible iterators" must
-also be SYCL device-copyable, as defined by the SYCL Specification.
+"Indirectly device accessible" types which are iterators must also be SYCL device-copyable to be used as input or output
+for oneDPL algorithms using a ``device_policy``.
 
 oneDPL Iterators
 ++++++++++++++++
@@ -201,6 +204,9 @@ to index into the value set defined by the source iterator. The resulting lvalue
 as the result of the operator.
 
 ``permutation_iterator`` is SYCL device-copyable if both the ``SourceIterator`` and the ``IndexMap``
+are SYCL device-copyable. ``permutation_iterator`` is "indirectly device accessible" if both the
+``SourceIterator`` and the ``IndexMap`` are "indirectly device accessible".
+
 are SYCL device-copyable. ``permutation_iterator`` is an "indirectly device accessible iterator" if both the
 ``SourceIterator`` and the ``IndexMap`` are "indirectly device accessible iterators".
 
@@ -326,8 +332,8 @@ source iterators over which the ``zip_iterator`` is defined. The arithmetic oper
 operation were applied to each of these iterators. The types ``T`` within the template pack 
 ``Iterators...`` must satisfy ``AdaptingIteratorSource``.
 
-``zip_iterator`` is SYCL device-copyable if all the source iterators are SYCL device-copyable, and
-is an "indirectly device accessible iterator" if all the source iterators are "indirectly device accessible iterators".
+``zip_iterator`` is SYCL device-copyable if all the source iterators are SYCL device-copyable, and is an "indirectly
+device accessible iterator" if all the source iterators are "indirectly device accessible iterators".
 
 .. code:: cpp
 
@@ -359,22 +365,22 @@ defining an Argument-Dependent Lookup (ADL) based customization point, ``is_oned
 oneDPL also defines a public trait, ``is_indirectly_device_accessible[_v]`` to indicate whether an iterator is an
 "indirectly device accessible iterators".
 
-oneDPL queries this information at compile time to determine how to handle iterator types when they are passed to
-algorithms with a ``device_policy`` to avoid unnecessary data movement.
+oneDPL queries this information at compile time to determine how to handle input and output types when they are passed
+to oneDPL algorithms with a ``device_policy`` to avoid unnecessary data movement.
 
 ADL Customization Point: ``is_onedpl_indirectly_device_accessible``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-A free function ``is_onedpl_indirectly_device_accessible(IteratorT)`` may be defined, which accepts an argument
-of type ``IteratorT`` and returns a type with the base characteristic of ``std::true_type`` if ``IteratorT`` refers to
-"indirectly device accessible", or otherwise returns a type with the base characteristic of ``std::false_type``. The
-function must be discoverable by ADL.
+A free function ``is_onedpl_indirectly_device_accessible(T)`` may be defined, which accepts an argument of type ``T``
+and returns a type with the base characteristic of ``std::true_type`` if ``T`` represents "indirectly device accessible"
+content, or otherwise returns a type with the base characteristic of ``std::false_type``. The function must be
+discoverable by ADL.
 
-The function ``is_onedpl_indirectly_device_accessible`` may be used by oneDPL to determine if the iterator refers
-to "indirectly device accessible" by interrogating its return type at compile-time only. Overloads may be provided as
-forward declarations only, without a body defined. ADL is used to determine which function overload to use according to
-the rules in the `C++ Standard`_. Therefore, derived iterator types without an overload for their exact type will match
-their most specific base iterator type if such an overload exists.
+The function ``is_onedpl_indirectly_device_accessible`` may be used by oneDPL to determine if the type represents
+"indirectly device accessible" content by interrogating its return type at compile-time only. Overloads may be provided
+as forward declarations only, without a body defined. ADL is used to determine which function overload to use according
+to the rules in the `C++ Standard`_. Therefore, derived iterator types without an overload for their exact type will
+match their most specific base iterator type if such an overload exists.
 
 Public Trait: ``oneapi::dpl::is_indirectly_device_accessible[_v]``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -391,8 +397,8 @@ The following class template and variable template are defined in ``<oneapi/dpl/
     inline constexpr bool is_indirectly_device_accessible_v = is_indirectly_device_accessible<T>::value;
 
 ``template <typename T> oneapi::dpl::is_indirectly_device_accessible`` is a type which has the base characteristic
-of ``std::true_type`` if ``T``  refers to "indirectly device accessible", otherwise it has the base characteristic of
-``std::false_type``.
+of ``std::true_type`` if ``T`` represents "indirectly device accessible" content, otherwise it has the base
+characteristic of ``std::false_type``.
 
 
 
@@ -411,7 +417,7 @@ of the iterator class.
     {
         struct accessible_it
         {
-            /* unspecified user definition of a iterator which refers to "indirectly device accessible" */
+            /* unspecified user definition of a iterator which represents "indirectly device accessible" content*/
         };
 
         std::true_type
@@ -419,7 +425,7 @@ of the iterator class.
 
         struct inaccessible_it
         {
-            /* unspecified user definition of iterator which doesn't refer to "indirectly device accessible" */
+            /* unspecified user definition of iterator which doesn't represent "indirectly device accessible" content */
         };
 
         // This could be omitted. It would rely upon the default implementation with the same result
--- a/source/elements/oneDPL/source/parallel_api/iterators.rst
+++ b/source/elements/oneDPL/source/parallel_api/iterators.rst
@@ -306,26 +306,29 @@ using the set of source iterators provided.
 
 .. _iterators-device-accessible:
 
-Iterators of Device Accessible Content with Device Policies
------------------------------------------------------------
+Device Accessible Content Iterators with Device Policies
+--------------------------------------------------------
 
 Iterator are, by default, not assumed to refer to content which is accessible on the device which requires the content
-to be copied to the device prior to being used inside a `SYCL`_ kernel. We say that iterators are of "device accessible
-content" when they can inherently be dereferenced on the device in a SYCL kernel. When using algorithms with a 
-``device_policy``, iterators of "device accessible content" avoid unnecessary data movement when provided as input or
-output arguments.
+to be copied to the device prior to being used inside a `SYCL`_ kernel. We define that iterators to be  "device
+accessible content iterators" when they can inherently be dereferenced on the device in a SYCL kernel. When using oneDPL
+algorithms with a ``device_policy``, "device accessible content iterators" avoid unnecessary data movement when provided
+as input or output arguments.
 
-Examples of iterators of "device accessible content" include SYCL USM shared or device memory, or iterator types like
+Examples of "device accessible content iterators" include SYCL USM shared or device memory, or iterator types like
 ``counting_iterator`` or ``discard_iterator`` that do not require any data to be copied to the device to be used in a
 SYCL kernel. An example of an iterator which does not refer to "device accessible content" is a ``std::vector``
 iterator, which requires the data to be copied to the device in some way prior to usage in a SYCL kernel within
 algorithms used with a ``device_policy``.
 
-oneDPL provides a mechanism to define whether custom iterator types are of "device accessible content" to SYCL kernels.
+oneDPL provides a mechanism to define whether custom iterators are "device accessible content iterators" by defining
+a Argument-Dependent Lookup (ADL) customization point, ``is_onedpl_device_accessible_content_iterator``. This allows
+users to define rules for their own iterators to be "device accessible content iterators" or not. oneDPL also defines a
+public trait, ``is_device_accessible_content_iterator[_v]`` to indicate whether an iterator is a "device accessible
+content iterator".
+
 oneDPL queries this information at compile time to determine how to handle the iterator type when passed to algorithms
-with a ``device_policy``. This is important because it allows oneDPL to avoid unnecessary data movement. This is
-achieved using the ``is_onedpl_device_accessible_content_iterator`` Argument-Dependent Lookup (ADL) customization point
-and the public trait ``is_device_accessible_content_iterator[_v]``.
+with a ``device_policy``. This is important because it allows oneDPL to avoid unnecessary data movement.
 
 ADL Customization Point: ``is_onedpl_device_accessible_content_iterator``
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
@@ -343,28 +346,29 @@ declarations only, without a body defined. ADL lookup is used to determine which
 the rules in the `C++ Standard`_. Therefore, derived iterator types without an overload for their exact type will match
 their most specific base iterator type if such an overload exists.
 
-The default implementation of ``is_onedpl_device_accessible_content_iterator`` marks the following iterators referring
-to "device accessible content":
+The default implementation of ``is_onedpl_device_accessible_content_iterator`` marks the following iterators as to
+"device accessible content iterators":
 * Pointers (to handle USM pointers)
 * Iterators with the ``using is_passed_directly = std::true_type`` trait
 * Iterators to USM shared allocated ``std::vector``-s when the allocator type is knowable from the iterator type
 * ``std::reverse_iterator<IteratorT>`` when ``IteratorT`` refers to "device accessible content"
 
-oneDPL defines its provided iterators as referring to "device accessible content" as follows:
-* ``counting_iterator`` and ``discard_iterator``: Always refer to "device accessible content".
-* ``permutation_iterator``:  Refers to "device accessible content" if both its source iterator and its index map refer
-to "device accessible content".
-* ``transform_iterator``:  Refers to "device accessible content" if its source iterator refers to "device accessible
-content".
-* ``zip_iterator``:  Refers to "device accessible content" if all base iterators refer to to "device accessible
-content".
+oneDPL defines rules for the provided iterators as follows:
+* ``counting_iterator``: Always a "device accessible content iterator"
+* ``discard_iterator``: Always a "device accessible content iterator"
+* ``permutation_iterator``:  "Device accessible content iterator" if both its source iterator and its index map are 
+"device accessible content iterators"
+* ``transform_iterator``:  "Device accessible content iterator" if its source iterator is a "device accessible content
+iterator"
+* ``zip_iterator``:  "Device accessible content iterator" if all base iterators are "device accessible content
+iterators"
 
 
 Public Trait: ``oneapi::dpl::is_device_accessible_content_iterator[_v]``
 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
 The public trait ``oneapi::dpl::is_device_accessible_content_iterator[_v]`` can be used to query whether an iterator
-refers to "device accessible content". The trait is defined in ``<oneapi/dpl/iterator>``.
+is a "device accessible content iterator". The trait is defined in ``<oneapi/dpl/iterator>``.
 
 ``oneapi::dpl::is_device_accessible_content_iterator<T>`` evaluates to a type with the characteristics of
 ``std::true_type`` if ``T``  refers to "device accessible content", otherwise it evaluates to a type with the
@@ -376,9 +380,9 @@ refers to "device accessible content", otherwise it evaluates to ``false``.
 Device Copyability of Iterators
 +++++++++++++++++++++++++++++++
 
-Iterators that refer to "device accessible content" and are used as input or output with algorithms using a
-``device_policy`` must also be SYCL device-copyable, as defined by the SYCL Specification. oneDPL provides
-specializations of ``sycl::is_device_copyable`` for each of its provided iterator types.
+"Device accessible content iterators" which are used as input or output with algorithms using a ``device_policy`` must
+also be SYCL device-copyable, as defined by the SYCL Specification. oneDPL provides specializations of
+``sycl::is_device_copyable`` for each of its provided iterator types.
 
 The rules are as follows:
 * ``counting_iterator``: Always device-copyable.
--- a/source/elements/oneDPL/source/parallel_api/iterators.rst
+++ b/source/elements/oneDPL/source/parallel_api/iterators.rst
@@ -360,14 +360,14 @@ device accessible* if all the source iterators are *indirectly device accessible
 ``make_zip_iterator`` constructs and returns an instance of ``zip_iterator``
 using the set of source iterators provided.
 
-Other Supported Iterators
-+++++++++++++++++++++++++
+Other Iterators
++++++++++++++++
+Pointers are assumed to be USM shared or device memory pointers and are *indirectly device accessible*.
+Pointers are trivially copyable and therefore SYCL device-copyable.
+
 ``std::reverse_iterator<IteratorT>`` is an ``AdaptingIteratorSource`` if ``IteratorT`` is an ``AdaptingIteratorSource``.
 ``std::reverse_iterator<IteratorT>`` is an *indirectly device accessible iterator* if ``IteratorT`` is an *indirectly
 device accessible iterator*.
-
-Pointers are assumed to be USM shared or device memory pointers and are *indirectly device accessible*.
-Pointers are trivially copyable and therefore SYCL device-copyable.
 
 Whether or not ``std::vector::iterator`` are *indirectly device accessible* is implementation defined, and may depend on
 the allocator used to create the vector, as well as the specific implementation of ``std::vector``.
--- a/source/elements/oneDPL/source/parallel_api/iterators.rst
+++ b/source/elements/oneDPL/source/parallel_api/iterators.rst
@@ -389,9 +389,9 @@ The following class template and variable template are defined in ``<oneapi/dpl/
     inline constexpr bool is_device_accessible_content_iterator_v =
         is_device_accessible_content_iterator<T>::value;
 
-``oneapi::dpl::is_device_accessible_content_iterator<T>`` evaluates to a type with the characteristics of
-``std::true_type`` if ``T``  refers to "device accessible content", otherwise it evaluates to a type with the
-characteristics of ``std::false_type``.
+``template <typename T> oneapi::dpl::is_device_accessible_content_iterator<T>`` evaluates to a type with the 
+characteristics of ``std::true_type`` if ``T``  refers to "device accessible content", otherwise it evaluates to a type
+with the characteristics of ``std::false_type``.
 
 
 
--- a/source/elements/oneDPL/source/parallel_api/iterators.rst
+++ b/source/elements/oneDPL/source/parallel_api/iterators.rst
@@ -207,8 +207,8 @@ as the result of the operator.
 are SYCL device-copyable. ``permutation_iterator`` is "indirectly device accessible" if both the
 ``SourceIterator`` and the ``IndexMap`` are "indirectly device accessible".
 
-are SYCL device-copyable. ``permutation_iterator`` is an "indirectly device accessible iterator" if both the
-``SourceIterator`` and the ``IndexMap`` are "indirectly device accessible iterators".
+When using ``permutation_iterator`` in combination with an algorithm with a ``device_policy``, ``SourceIterator`` must
+be "indirectly device accessible".
 
 .. code:: cpp
 
--- a/source/elements/oneDPL/source/parallel_api/iterators.rst
+++ b/source/elements/oneDPL/source/parallel_api/iterators.rst
@@ -369,7 +369,7 @@ Other Supported Iterators
 ``std::reverse_iterator<IteratorT>`` is an *indirectly device accessible iterator* if ``IteratorT`` is an *indirectly
 device accessible iterator*. The SYCL device-copyable requirement of ``std::reverse_iterator<IteratorT>`` for use in
 algorithms with a ``device_policy`` relies upon the trivial copyability of ``IteratorT`` and the specific implementation
-of ``std::reverse_iterator``. oneDPL does not specialize ``sycl::is_device_copyable`` for ``std::reverse_iterator``.
+of ``std::reverse_iterator``. 
 
 Pointers are assumed to be USM shared or device memory pointers and are *indirectly device accessible*.
 Pointers are trivially copyable and therefore SYCL device-copyable.
--- a/source/elements/oneDPL/source/parallel_api/iterators.rst
+++ b/source/elements/oneDPL/source/parallel_api/iterators.rst
@@ -397,7 +397,7 @@ source iterator component types. Refer to the example below.
 Example
 ^^^^^^^
 
-The following example shows how to define a customization for `is_indirectly_device_accessible` trait for a simple
+The following example shows how to define a customization for ``is_indirectly_device_accessible`` trait for a simple
 user defined iterator.  It also shows a more complicated example where the customization is defined as a hidden friend
 of the iterator class.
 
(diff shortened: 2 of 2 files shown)


source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +303,116 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+should be "passed directly" to SYCL kernels by interrogating its return type at compile time only. It shall not be
+called by oneDPL outside a ``decltype`` context to determine the return type. This means that overloads may be provided
+as forward declarations only, without a body defined. ADL lookup is used to determine which function overload to use
+according to the rules in the C++ Standard. Therefore, derived iterator types without an overload for their exact type
+will match their most specific base iterator type if such an overload exists.
+
+The default implementation of ``is_passed_directly_in_onedpl_device_policies`` marks the following iterators as
+"passed directly":
+* Pointers (to handle USM pointers)
+* Iterators with the ``using is_passed_directly = std::true_type`` trait
```
danhoeflinger, line 341: We likely want to remove this from the official specification, even if we choose to include this in our implementation (possibly temporarily).  
Including this would be to support any existing code relying upon an existing unspecified feature.  Any code relying on this can instead update to rely upon this specified feature.

source/elements/oneDPL/source/parallel_api/execution_policies.rst
```diff
@@ -99,7 +99,11 @@ to run algorithms on a SYCL device. When an algorithm runs with ``device_policy`
 it is capable of processing SYCL buffers (passed via ``oneapi::dpl::begin/end``),
 data in the host memory and data in Unified Shared Memory (USM), including device USM.
 Data placed in the host memory and USM can be passed to oneDPL algorithms
-as pointers and random access iterators. The way to transfer data from the host memory
+as pointers and random access iterators. oneDPL provides some :ref:`iterators <iterators>` that are
+compatible with algorithms when using a ``device_policy``. Custom iterators may also be used, but users should
+ensure that those iterators have a defined :ref:`"passed directly" customization point <iterators-passed-directly>`
+to avoid unnecessary data movement. The iterators must also be SYCL device-copyable to be used with
+algorithms utilizing a ``device_policy``. The way to transfer data from the host memory
```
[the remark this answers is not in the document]
danhoeflinger: I'm not incredibly happy with the current language I have here, but I think *something* should be here in this section because it is already discussing the valid options for input data when using a device policy. Perhaps just a brief link to the iterators page, and then most of this can go at the top of that page instead?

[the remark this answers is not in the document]
akukanov: It discusses primarily the placement of data for device policies, not the types of iterators. You can create fancy iterators on top of host containers - but that would not make the data suitable to process on a device.

I am open for some short reference to the iterators, but it should be properly put into the context.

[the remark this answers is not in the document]
danhoeflinger: Ive made this change locally, will push today.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +303,109 @@ operation were applied to each of these iterators. The types ``T`` within the te
 
 ``make_zip_iterator`` constructs and returns an instance of ``zip_iterator``
 using the set of source iterators provided.
+
+.. _iterators-passed-directly:
+
+Customization Point for "Passed Directly" Iterators
```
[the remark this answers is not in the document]
danhoeflinger: My current local draft uses iterators referring to "device accessible content".  I think its an improvement in clarity (though slightly more verbose).  I'm working out some details, but will push today.

content rather than data to include things like `counting_iterator` / `discard_iterator`.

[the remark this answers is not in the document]
danhoeflinger: See if the newest changes help with this.  I think we can tweak the language a bit as necessary, but this frames it based upon the characteristics of the iterator / data rather than the implementation detail of what does or doesn't happen as a result.

source/elements/oneDPL/source/parallel_api/execution_policies.rst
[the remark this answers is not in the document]
danhoeflinger: Shortened this to just provide a link.  Added some language about device copyability to the iterators page.

source/elements/oneDPL/source/parallel_api/execution_policies.rst
```diff
@@ -99,8 +99,8 @@ to run algorithms on a SYCL device. When an algorithm runs with ``device_policy`
 it is capable of processing SYCL buffers (passed via ``oneapi::dpl::begin/end``),
 data in the host memory and data in Unified Shared Memory (USM), including device USM.
 Data placed in the host memory and USM can be passed to oneDPL algorithms
-as pointers and random access iterators. The way to transfer data from the host memory
-to a device and back is unspecified; per-element data movement to/from a temporary storage
+as pointers and random access iterators, please see :ref:`iterators <iterators>`. The way to transfer data from the host
+memory to a device and back is unspecified; per-element data movement to/from a temporary storage
 is a possible valid implementation.
```
akukanov, line 104:
```suggestion
as pointers and random access iterators, including :doc:`oneDPL iterators <iterators>`. The way 
to transfer data from the host memory to a device and back is unspecified; per-element data
movement to/from a temporary storage is a possible valid implementation.
```

source/elements/oneDPL/source/parallel_api/iterators.rst
akukanov, line 1: The new wording is pretty good overall, but it seems the whole page should be structured differently to make more sense. I see roughly the following structure (the final names can be different):

- Iterators (the main page header)
  - Requirements to use with device policies
    Here we describe the semantical requirements for device accessible content (without the trait and the customization) as well as device copyability
  - Special oneDPL iterators
    - The AdaptingIteratorSource requirement
    - counting_iterator
    - discard_iterator
    - etc.
      The specific conditions where the special iterators are suitable for device policies belong to these chapters, rather than separate lists.
  - Customization for user defined iterators

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: I've removed this.

source/elements/oneDPL/source/parallel_api/execution_policies.rst
danhoeflinger: suggestion taken.

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: I adjusted the structure accordingly, and distributed the rules to be with the types.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +339,105 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+
+
+
+
+Example
+^^^^^^^
+
+The following example shows how to define a ADL overload for a simple user defined iterator.  It also shows an example
+for a more complicated case which uses a hidden friend to provide the ADL overload within the definition of the
+iterator.
```
rarutyun, line 401:
```suggestion
The following example shows how to define a customization for `is_device_accessible_content_iterator` trait for a simple user defined iterator.  It also shows a more complicated example where the customization is defined as a hidden friend of the iterator class.
```

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +339,105 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+
+The default implementation of ``is_onedpl_device_accessible_content_iterator`` marks the following iterators as to
+"device accessible content iterators":
+* Pointers (to handle USM pointers)
+* Iterators to USM shared allocated ``std::vector``-s when the allocator type is knowable from the iterator type
+* ``std::reverse_iterator<IteratorT>`` when ``IteratorT`` refers to "device accessible content"
+
+
+
+Public Trait: ``oneapi::dpl::is_device_accessible_content_iterator[_v]``
```
rarutyun, line 380: naming suggestions to consider

- `is_passed_directly_to_device` (perhaps obsolete)
- `is_device_accessible_iterator`
- `is_indirectly_device_accessible`

For the last one it might be `is_device_indirectly_accessible`; not sure what is more accurate from English perspective. But the last proposed option (by me) is my favorite, so far. It also corelates with several C++20 concepts, e.g., `indirectly_writable`

rarutyun (commented):
Despite a significant amount of suggestions, I think that wording is pretty good, overall. There is one place that is related to the design, where we need to be careful to not prohibit some intendent use-cases.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +339,105 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+to "device accessible content" by interrogating its return type at compile-time only. It shall not be called by oneDPL
+outside a ``decltype`` context to determine the return type. Overloads may be provided as forward declarations only,
+without a body defined. ADL lookup is used to determine which function overload to use according to the rules in the
+`C++ Standard`_. Therefore, derived iterator types without an overload for their exact type will match their most
+specific base iterator type if such an overload exists.
+
+The default implementation of ``is_onedpl_device_accessible_content_iterator`` marks the following iterators as to
+"device accessible content iterators":
+* Pointers (to handle USM pointers)
+* Iterators to USM shared allocated ``std::vector``-s when the allocator type is knowable from the iterator type
```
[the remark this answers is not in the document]
danhoeflinger: In practice, with a good amount of certainty with some compilers, yes. However, its a good question if its worth mentioning in the specification, since it is not always knowable.
https://github.com/uxlfoundation/oneDPL/pull/1438#issuecomment-1989312899
https://github.com/uxlfoundation/oneDPL/pull/1438#issuecomment-2032850867

How it looks in the oneDPL's implementation currently:
https://github.com/uxlfoundation/oneDPL/blob/1625f6a2dcc981a537c65fdc3b40951cb63b7326/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h#L154-L179

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: I don't really like `is_device_accessible_iterator` because I think there is a difference between the iterator being accessible and its content it refers to being accessible.  However, `is_indirectly_device_accessible` is more clear and concise than what is currently in the PR.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +339,105 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+
+
+
+Public Trait: ``oneapi::dpl::is_device_accessible_content_iterator[_v]``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The public trait ``oneapi::dpl::is_device_accessible_content_iterator[_v]`` can be used to query whether an iterator
+is a "device accessible content iterator". The trait is defined in ``<oneapi/dpl/iterator>``.
+
+``oneapi::dpl::is_device_accessible_content_iterator<T>`` evaluates to a type with the characteristics of
```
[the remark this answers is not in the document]
danhoeflinger: Adjusted to be close to your proposal.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -174,6 +200,10 @@ to index into the index map. The corresponding value in the map is then used
 to index into the value set defined by the source iterator. The resulting lvalue is returned
 as the result of the operator.
 
+``permutation_iterator`` are SYCL device-copyable if both the ``SourceIterator`` and the ``IndexMap``
+are SYCL device-copyable, and are always "device accessible content iterators" if both the ``SourceIterator``
```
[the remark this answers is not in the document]
danhoeflinger: removed all instances of this.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +339,105 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+A free function ``is_onedpl_device_accessible_content_iterator(IteratorT)`` may be defined, which accepts an argument
+of type ``IteratorT`` and returns a type with the characteristics of ``std::true_type`` if ``IteratorT`` refers to
+"device accessible content", or alternatively returns a type with the characteristics of ``std::false_type``
+otherwise. The function must be defined in one of the valid search locations for ADL lookup, which includes the
+namespace of the definition of the iterator type ``IteratorT``.
+
+The function ``is_onedpl_device_accessible_content_iterator`` may be used by oneDPL to determine if the iterator refers
+to "device accessible content" by interrogating its return type at compile-time only. It shall not be called by oneDPL
+outside a ``decltype`` context to determine the return type. Overloads may be provided as forward declarations only,
+without a body defined. ADL lookup is used to determine which function overload to use according to the rules in the
```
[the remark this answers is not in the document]
danhoeflinger: taken

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +339,105 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The public trait ``oneapi::dpl::is_device_accessible_content_iterator[_v]`` can be used to query whether an iterator
+is a "device accessible content iterator". The trait is defined in ``<oneapi/dpl/iterator>``.
+
+``oneapi::dpl::is_device_accessible_content_iterator<T>`` evaluates to a type with the characteristics of
+``std::true_type`` if ``T``  refers to "device accessible content", otherwise it evaluates to a type with the
+characteristics of ``std::false_type``.
+
+``oneapi::dpl::is_device_accessible_content_iterator<T>`` is a ``constexpr bool`` that evaluates to ``true`` if ``T``
```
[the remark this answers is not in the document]
danhoeflinger: Rearranged so that this description is just in the code.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +339,105 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+
+oneDPL queries this information at compile time to determine how to handle iterator types when they are passed to
+algorithms with a ``device_policy`` to avoid unnecessary data movement.
+
+ADL Customization Point: ``is_onedpl_device_accessible_content_iterator``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A free function ``is_onedpl_device_accessible_content_iterator(IteratorT)`` may be defined, which accepts an argument
+of type ``IteratorT`` and returns a type with the characteristics of ``std::true_type`` if ``IteratorT`` refers to
+"device accessible content", or alternatively returns a type with the characteristics of ``std::false_type``
```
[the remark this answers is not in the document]
danhoeflinger: taken.

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: taken.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +339,105 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+otherwise. The function must be defined in one of the valid search locations for ADL lookup, which includes the
+namespace of the definition of the iterator type ``IteratorT``.
```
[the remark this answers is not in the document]
danhoeflinger: Taken, I removed the second part.  It doesn't really make sense after the edit.  Its more something for the documentation than the spec.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +339,105 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+accessible content iterator".
+
+oneDPL queries this information at compile time to determine how to handle iterator types when they are passed to
+algorithms with a ``device_policy`` to avoid unnecessary data movement.
+
+ADL Customization Point: ``is_onedpl_device_accessible_content_iterator``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A free function ``is_onedpl_device_accessible_content_iterator(IteratorT)`` may be defined, which accepts an argument
+of type ``IteratorT`` and returns a type with the characteristics of ``std::true_type`` if ``IteratorT`` refers to
```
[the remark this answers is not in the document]
akukanov: > However, std::conjunction might not have `std::true_type` or `std::false_type` characteristics at all.

If it is used with `true_type`/`false_type` or derivatives, it will:
> The specialization `std::conjunction<B1, ..., BN>` has a public and unambiguous base that is
> - if `sizeof...(B) == 0`, `std::true_type`;
> - otherwise the first type Bi in B1, ..., BN for which `bool(Bi::value) == false`, or BN if there is no such type.

 And that's exactly our goal I believe.

source/elements/oneDPL/source/parallel_api/iterators.rst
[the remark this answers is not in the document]
akukanov: Think of that as of specification, not as of documentation.
If we want to require that `std::vector` with USM allocators is used without excessive copying, then it is a part of the spec. Otherwise, it can be mentioned as an implementation-specific or completely omitted.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +339,105 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+"device accessible content", or alternatively returns a type with the characteristics of ``std::false_type``
+otherwise. The function must be defined in one of the valid search locations for ADL lookup, which includes the
+namespace of the definition of the iterator type ``IteratorT``.
+
+The function ``is_onedpl_device_accessible_content_iterator`` may be used by oneDPL to determine if the iterator refers
+to "device accessible content" by interrogating its return type at compile-time only. It shall not be called by oneDPL
+outside a ``decltype`` context to determine the return type. Overloads may be provided as forward declarations only,
```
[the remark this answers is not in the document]
danhoeflinger: Well, it justifies the technical reason that "Overloads may be provided as forward declarations only,...", but perhaps its just enough to say that.  I've removed it.

source/elements/oneDPL/source/parallel_api/iterators.rst
akukanov: I am fine with `is_indirectly_device_accessible`, but we need to get feedback from the customers.

source/elements/oneDPL/source/parallel_api/iterators.rst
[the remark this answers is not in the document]
danhoeflinger: In practice, oneDPL's implementation is relying upon standard library implementation details which are not part of the C++ standard library specification of `std::vector::iterator` to enable handling of usm allocator vector iterators when it can confidently guess that the allocator type is knowable from the iterator type.  From purely the C++ standard specification, I don't think it is possible to definitively know anything about the allocator from the type of a `std::vector::iterator`. 

In practice, it would be quite unlikely that a standard library implementation could / would have a `std::vector::iterator` type which "tricks" the current oneDPL implementation's detection of a knowable "USM shared allocator", but I do think you could technically contrive such an implementation if you really wanted to.

From a specification perspective, it doesn't belong, I think.

source/elements/oneDPL/source/parallel_api/iterators.rst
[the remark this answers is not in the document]
danhoeflinger: I suppose I don't know exactly what we mean by "characteristics of `std::true_type`" if a `std::conjunction` or similar types will not qualify.

source/elements/oneDPL/source/parallel_api/iterators.rst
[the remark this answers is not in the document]
danhoeflinger: I've removed the line for now, but I'm open to discussion of course.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -64,6 +86,8 @@ counter; dereference operations cannot be used to modify the counter. The arithm
 operators of ``counting_iterator`` behave as if applied to the values of Integral type
 representing the counters of the iterator instances passed to the operators.
 
+``counting_iterator`` are SYCL device-copyable, and are "device accessible content iterators".
```
rarutyun, line 89: I was fine with keeping "always for `counting_iterator` and `discard_iterator` (I think I wrote something like "iterator below". But how it's right now is also good. Please keep it as is.

I have another suggestion somewhere you use "`<iterator_name>` are" and somewhere "`<iterator_name>` is". I propose to align to "is" everywhere
```suggestion
``counting_iterator`` is SYCL device-copyable, and are "device accessible content iterators".
```

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -104,6 +128,8 @@ lvalue that may be assigned an arbitrary value. The assignment has no effect on
 of ``discard_iterator`` behave as if applied to integer counter values maintained by the
 iterator instances to determine their position relative to each other.
 
+``discard_iterator`` are SYCL device-copyable, and are "device accessible content iterators".
```
rarutyun, line 131:
```suggestion
``discard_iterator`` is SYCL device-copyable, and are "device accessible content iterators".
```

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -174,6 +200,10 @@ to index into the index map. The corresponding value in the map is then used
 to index into the value set defined by the source iterator. The resulting lvalue is returned
 as the result of the operator.
 
+``permutation_iterator`` are SYCL device-copyable if both the ``SourceIterator`` and the ``IndexMap``
```
rarutyun, line 203:
```suggestion
``permutation_iterator`` is SYCL device-copyable if both the ``SourceIterator`` and the ``IndexMap``
```

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +339,108 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+
+    template <typename T>
+    struct is_device_accessible_content_iterator{ /* see below */ };
+
+    template <typename T>
+    inline constexpr bool is_device_accessible_content_iterator_v = is_device_accessible_content_iterator<T>::value;
+
+``template <typename T> oneapi::dpl::is_device_accessible_content_iterator<T>`` evaluates to a type with the 
+characteristics of ``std::true_type`` if ``T``  refers to "device accessible content", otherwise it evaluates to a type
+with the characteristics of ``std::false_type``.
```
rarutyun, line 393: I would still recommend some improvements
```suggestion
``is_device_accessible_content_iterator`` has characteristics of ``std::true_type`` if ``T``  refers to "device accessible content", otherwise it has characteristics of ``std::false_type``.
```
If you want to keep `template <typename T>` at the beginning, that's probably fine (though, looks redundant) but please at least remove `<T>` after class name. Also, "evaluates to a type" sounds a little bit clunky to me.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +339,108 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+user defined iterator.  It also shows a more complicated example where the customization is defined as a hidden friend
+of the iterator class.
+
+.. code:: cpp
+
+    namespace usr
+    {
+        struct accessible_it
+        {
+        /* unspecified user definition of a iterator which refers to "device accessible content" */
```
rarutyun, line 412:
```suggestion
            /* unspecified user definition of a iterator which refers to "device accessible content" */
```

source/elements/oneDPL/source/parallel_api/iterators.rst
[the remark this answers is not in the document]
rarutyun: Yes, but the thing about `decltype` is that it makes allowed use-cases broader. So, users are allowed to provide both declarations and `constexpr` definitions. If it would narrow the use-cases It would have made a perfect sense to me.

With the current design I am glad that you decided to remove it.

source/elements/oneDPL/source/parallel_api/iterators.rst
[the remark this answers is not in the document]
rarutyun: Yeah, if we speak about `Allocator` we should write an iterator type slightly differently: `std::vector<T, Allocator>::iterator`.

As you can see `Allocator` is a part of specialized `vector` type, not an `iterator` type. I don't know a way in C++ do deduce a outer type by inner type: in this concrete example: decude `vector<T, Allocator>` type from nested `iterator` type

source/elements/oneDPL/source/parallel_api/iterators.rst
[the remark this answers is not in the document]
danhoeflinger: I don't think I understand this most recent comment. As it stands, users are able to provide either declarations or  full definitions.  As we discussed offline previously, return type must be knowable at compile-time regardless of what is marked as `constexpr` or not, so any implementation for which that isn't the case will have errors anyway.

I'm not understanding why knowing the `decltype` thing actually changes anything other than "Overloads may be provided as forward declarations only...", other than giving some insight into why that is the case technically.

Can you help me understand with an example of something allowed for a user in with this `decltype` knowledge which isn't allowed with the current language?

source/elements/oneDPL/source/parallel_api/iterators.rst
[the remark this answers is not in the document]
akukanov: > I suppose I don't know exactly what we mean by "characteristics of std::true_type" if a std::conjunction or similar types will not qualify.

The wording "with a base characteristic of" is used in a few places in the C++ standard, for example in https://eel.is/c++draft/meta.unary:
>Subclause [meta.unary] contains templates that may be used to query the properties of a type at compile time.
> Each of these templates shall be a Cpp17UnaryTypeTrait **with a base characteristic of true_type if the corresponding condition is true, otherwise false_type**.

My understanding of this wording is like: even though it is not a true_type or false_type, it has all the same characteristics as these types and can be used as such. Practically it most often means that the trait class inherits either true_type or false_type (maybe indirectly).

In my comment above, I say that std::conjunction will also fall into this category if the chosen template parameter does that. And though the standard does not prohibit to use other types with `std::conjunction`, the typical arguments it is used with will be traits etc. classes that inherit `bool_constant`. Therefore requiring that the return type of our ADL customization function should have these characteristics does not preclude the use of `std::conjunction`, but limits what types can be passed to it.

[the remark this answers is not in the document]
danhoeflinger: Is see, thanks for the clarification.  I think I agree that our current formulation makes sense then, with a small tweak to the language perhaps to follow the example language you provided. 

Unless we have a compelling example to allow `conjunction` `disjunction` `negation` which use something other than types with the base characteristic of a `bool_constant`, I think simplicity is better.

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: adjusted all of these "are" -> "is" [a]

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: adjusted all of these "are" -> "is" [a]

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: adjusted all of these "are" -> "is" [a]

danhoeflinger: adjusted all of these "are" -> "is" [a]

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: I adjusted the name here in the spec.

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: fixed.

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: fixed.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +337,108 @@ operation were applied to each of these iterators. The types ``T`` within the te
 
 ``make_zip_iterator`` constructs and returns an instance of ``zip_iterator``
 using the set of source iterators provided.
+
+.. _iterators-device-accessible:
+
+Customization For User Defined Iterators
```
akukanov, line 343: What should we do with standard iterator adaptor types that we (would like to) support? `reverse_iterator`, `const_iterator`, `move_iterator`, maybe `counted_iterator`, as well as the iterators for supported C++ view types?

We might say nothing in the spec and leave it to the implementation, or we might claim these types as conditionally device accessible, similar to our iterator adaptors.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -3,9 +3,31 @@
 ..
 .. SPDX-License-Identifier: CC-BY-4.0
 
+.. _iterators:
+
 Iterators
 ---------
 
+Requirements For Iterator Use With Device Policies
```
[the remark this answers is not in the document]
danhoeflinger: I've thought about this and I think the answer is "no" both both from a technical perspective and from a semantic one.

From a technical perspective, we are checking what can be passed in without any processing into SYCL kernels.  The return type of `oneapi::dpl::begin/end` requires processing to be passed into a SYCL kernel.  It is a valid input to oneDPL API, but it does require processing to convert it to an accessor during a submit call to a SYCL kernel.

From a semantic perspective, the simple answer is that it is not an iterator.  

We have to be careful not to confuse "valid inputs to oneDPL APIs" with "indirectly device accessible iterators". For instance, `std::vector::iterator` with host allocated data is also a valid input but it is not an "indirectly device accessible iterator". It is valid to use these return types of `oneapi::dpl::begin/end`  in combination with oneDPL's provided iterators, but they cannot be passed without processing into SYCL kernels.

Can the return types of `oneapi::dpl::begin/end` be used in combination with custom user iterators?  No, not without additional infrastructure, which is beyond the scope of feature I believe.

We probably need some language clarifying this...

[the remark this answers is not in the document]
danhoeflinger: Similarly, `std::vector::iterator` with host data can not be enabled for use with custom user iterators without additional infrastructure.  We would need some sort of customization point where users could provide information oneDPL would need to break an iterator down into sub-components which are all valid oneDPL API inputs.

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: `reverse_iterator` is covered both in the spec language in the default overload of the customization point and by the implementation.

Should we comment on `counted_iterator`(c++20) or `const_iterator`(c++23) with c++17 as our support baseline outside of the ranges API (for which this feature is not relevant)?

I don't have a good answer for `move_iterator` (or counted, const) other than they aren't currently supported by oneDPL's current implementation processing inputs in preparation for SYCL kernels (to the best of my knowledge).

source/elements/oneDPL/source/parallel_api/iterators.rst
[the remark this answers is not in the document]
akukanov: Ok, so it's a semantical question.

What we propose now is a "boolean" trait that for any given type can be either "true" or "false".
We say that for e.g. pointers it is "true" because we prefer to treat them as USM pointers and avoid excessive copying.
We say that for arbitrary (unknown) iterators it is "false" because we cannot guarantee the data being on the device, and conservatively prefer to take a copy.
We introduce a customizable trait for the sole reason to avoid the overhead of copying when users know that the data are accessible on the device.
Is this the semantics, or do I miss something?

If this is the semantics, then why for pseudo-iterators over the buffers the trait comes to "false"? They do not require any data copying, as SYCL implementation would take care of it. In my opinion, the semantics for buffers is closer to that of pointers than to that of unknown iterators.

source/elements/oneDPL/source/parallel_api/iterators.rst
akukanov: > reverse_iterator is covered both in the spec language in the default overload of the customization point and by the implementation.

It's covered in a somewhat strange way. To me, it is strange that the specification talks about "the default implementation". It might describe the required implementation for an arbitrary argument type (in our case, the default is set to "false_type") and required specializations for other types, such as `reverse_iterator`.

> Should we comment on counted_iterator(c++20) or const_iterator(c++23) with c++17 as our support baseline outside of the ranges API (for which this feature is not relevant)?

It's not about versions of the standard, it's about whether we want the specification to require certain behavior for these types (when they are available) or not. Or, putting it differently, C++17 is the minimal version of the standard but not the only one - so the standard APIs available in later versions still can be considered.

source/elements/oneDPL/source/parallel_api/iterators.rst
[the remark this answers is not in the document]
danhoeflinger: I suppose this is one reason why the original "passed directly" naming had some benefits in that it was describing what is useful for oneDPL's implementation to actually want to know:  "Is this able to be just passed into a SYCL kernel without processing?".  

It gets a little confused with the new naming, and I understand the want to classify the return types of `oneapi::dpl::begin/end` as "indirectly device accessible", but from a technical perspective this does not provide the answer to the real question that the oneDPL implementation can use when processing inputs.  

We can have some language here that says something like: 
"While the return type of `oneapi::dpl::begin/end` is not technically an "indirectly device accessible iterator", it does not incur extra overhead beyond the normal handling of a `sycl::buffer` by the SYCL runtime." 

or, we can move back to some other naming like "passed directly".  If we mark return type of `oneapi::dpl::begin/end` as `true` here, then it stops being useful from a technical perspective.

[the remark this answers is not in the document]
danhoeflinger: Also, if we look closely at how we already define "indirectly device accessible iterator" is says the iterator type "can inherently be dereferenced on the device in a SYCL kernel".  This is not true of return type of `oneapi::dpl::begin/end`.  We explicitly state in the spec that that the return type of `oneapi::dpl::begin/end`  "cannot be used as iterators in other contexts, including dereference".

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: We can change the way we describe this by saying the default is `std::false_type`, and then describe pointers and `reverse_iterators` and other iterator adaptors separately.

danhoeflinger: I suppose the question for `const_iterator`, `move_iterator`, maybe `counted_iterator` is whether we want to add mention of general support for these iterators in the context of this feature, or if we want to treat that as a separate change.  

From the specification perspective, I'm not sure. From the implementation perspective, it may either put at risk the implementation of this feature landing in the upcoming milestone or put us out of step with the specification if we defer that specific functionality for later. It seems like it shouldn't be difficult, but without more exploration, I can't say.

danhoeflinger: Ok, sounds good.  I can make an issue to explore support for these iterator adaptors in oneDPL.  I can also adjust the language to discuss `reverse_iterator` on its own rather than in the default implementation section.

source/elements/oneDPL/source/parallel_api/iterators.rst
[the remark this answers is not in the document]
akukanov: @danhoeflinger you look at it from the background of the current oneDPL implementation and its needs, while I am trying to look at it from a more abstract, usage-centric view point (in a broad sense of usage like a contract between implementors and users of the library). That's why I push to move away from the implementation-centric "passed directly" paradigm.

Whoever calls our algorithms, does not care how the "iterators" are passed onto the device and if they are dereferenced there or not. That is honestly implementation details, for as long as algorithms process the data. Whether the data are copied before processing and then back or not, however, can be, and I think should be, considered as a part of the contract, because it directly impacts performance and so usage experience.

Therefore it makes sense for me that the trait answers the question whether a oneDPL implementation will be able to avoid excessive data copying for my special iterators or not. But if that's not the question we want the trait to answer, then what is the real question, which would make sense as part of the library contract, and not an implementation detail?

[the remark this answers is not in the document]
danhoeflinger: I think the real question is: 
"Can this type be inherently be dereferenced on the device in a SYCL kernel"?

It just so happens that oneDPL makes use of this trait to avoid excessive data copying, but that is not what the trait explicitly means.  We could make a different trait if users are curious about if a type will require a copy into a `sycl::buffer` on its way into a SYCL kernel.  `template<typename Iter>requires_copy_to_sycl_buffer_in_device_policies[_v]` or something like that.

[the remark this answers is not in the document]
akukanov: Yet in the implementation we imply that if it cannot be dereferenced then we need to copy the data - except if it's the "buffer position type" which we cannot dereference but should not copy either. Does not sound like a proper implication to me.

The proper implications, I think, would be these:
If a type represents a data sequence/range that is either accessible on the device or generated there on the fly, then:
- the implementation should not take an extra step of data copying;
- if the type is also device copyable, it can be passed as the argument to a SYCL kernel;
  - otherwise, the implementation should somehow obtain a different representation of the data sequence/range that can be passed in.
- and if the type is also dereferenceable (including with operator[]), it can be used directly in the kernel;
  - otherwise, the implementation should access data in some different manner.

So do you say that the trait should only be true if all three conditions are satisfied, or the first and the last because for device copyability we have a different trait?

I think that it would be sufficient for the trait to answer just the first question, which for the user types we (the library implementors) cannot really answer in any other way. But we certainly can say whether a type is device copyable (by checking the corresponding SYCL trait) and we can say if it is dereferenceable (e.g. by simply requiring that it's a random access iterator, which by the standard is mandated to have `operator[]`).

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: I've added a section to describe characteristics of `std::reverse_iterator` and USM pointers. We can add to this section later.

However, I do think it still makes sense to describe them in the context of the default implementation.  These types won't have their own ADL overload because a pointer is a built-in type and without knowledge of `std::reverse_iterator` implementation details we can't create an ADL overload safely in the `std` namespace.

akukanov: Let's clear the terms. The default implementation of the trait or the default implementation of the library?
For the first, a specialization for a certain type is by definition not a default implementation.
For the second, in which sense a library implementation would be "default", especially from the specification standpoint? :)

danhoeflinger: Sorry, default overload for ADL base customization point.  At compile time it checks the type and sets its return type accordingly.

akukanov: The "default overload of the customization point" is an implementation detail in my opinion. Unless I miss something, such an overload does not have to exist.

danhoeflinger: With a separate section for these other iterators, I've removed the implementation details about how it would be implemented.  
You are correct, the important thing is that we set out the semantic rules, and whether we implement them as a default ADL overload or the specialization of a trait doesn't belong here.

source/elements/oneDPL/source/parallel_api/iterators.rst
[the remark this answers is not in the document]
danhoeflinger: To follow up here, I've adjusted the language slightly, but think this simple version is best without making it more broad. I've not been able to come up with any real compelling case for `conjunction` or `disjunction` of an `integral_constant` which is not a `bool_constant` worth making this language more complex.

However, I think a valid implementation could easily accommodate arbitrary usage of `conjunction` and `disjunction`, ensuring that the public trait is always a `bool_constant`.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -3,9 +3,36 @@
...
+SYCL kernel. An example of an iterator type that is not "indirectly device accessible" is a ``std::vector`` iterator
+with a host allocator, which requires the data to be copied to the device in some way before usage in a SYCL kernel
+within algorithms used with a ``device_policy``.
+
+:doc:`*buffer position objects* <buffer_wrappers>` returned by ``oneapi::dpl::begin`` and ``oneapi::dpl::end`` are not
+iterators, but they are "indirectly device accessible" because they represent data that is accessible on the device.
+When using oneDPL algorithms with a ``device_policy``, "indirectly device accessible" types avoid unnecessary data
+movement when provided as input or output arguments.
+
+"Indirectly device accessible iterators" must also be SYCL device-copyable to be used as input or output for oneDPL
```
mmichel11, line 30: Do we also want to explicitly require that they must be random access here? In the current proposed implementation this is a requirement.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -234,6 +272,9 @@ arithmetic and comparison operators of ``transform_iterator`` behave as if appli
 source iterator itself. The template type ``Iterator`` must satisfy
 ``AdaptingIteratorSource``.
 
+``transform_iterator`` is SYCL device-copyable if the source iterator is SYCL device-copyable, and
+is "indirectly device accessible iterator" if the source iterator is an "indirectly device accessible iterators".
```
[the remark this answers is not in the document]
danhoeflinger: fixed, as well as some others.

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: I changed this section, but I do now mention random access iterator specifically

danhoeflinger:
After some offline discussion with @akukanov, I did a significant rewrite of this spec to try to focus on requirements from the specification rather than implementation details of the existing oneDPL implementation, as well as some other changes.  Hopefully this better matches the expectations.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +362,88 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+
+        std::true_type
+        is_onedpl_indirectly_device_accessible(accessible_it);
+
+        struct inaccessible_it
+        {
+            /* unspecified user definition of iterator  which is not "indirectly device accessible" */
+        };
+
+        // The following could be omitted. The default overload would provide the same result.
```
akukanov, line 426:
```suggestion
        // The following could be omitted, as returning std::false_type matches the default behavior.
```

danhoeflinger:
@akukanov All good suggestions... applied.  Thanks.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -3,9 +3,53 @@
...
+Requirements For Iterator Use With Device Policies
+++++++++++++++++++++++++++++++++++++++++++++++++++
+Iterators and iterator-like types may or may not refer to content accessible on the device. The term *indirectly device
+accessible* refers to a type that represents content accessible on the device within a `SYCL`_ kernel. *Indirectly
+device accessible iterators* are iterators that can inherently be dereferenced on the device within a SYCL kernel.
+
+Examples of *indirectly device accessible iterators* include SYCL USM shared or device memory pointers, as well as
+iterator types like ``counting_iterator`` or ``discard_iterator``, which can inherently be used in a SYCL kernel. An
+example of an iterator type that is not *indirectly device accessible* is a ``std::vector`` iterator with a host
+allocator, as it cannot inherently be used in a SYCL kernel.
```
akukanov, line 20: These examples seem mostly redundant, because the same information is provided below. Perhaps one example would be enough to illustrate the introduced terms.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -3,9 +3,53 @@
...
+.. _iterators:
+
 Iterators
 ---------
 
+Requirements For Iterator Use With Device Policies
+++++++++++++++++++++++++++++++++++++++++++++++++++
+Iterators and iterator-like types may or may not refer to content accessible on the device. The term *indirectly device
+accessible* refers to a type that represents content accessible on the device within a `SYCL`_ kernel. *Indirectly
+device accessible iterators* are iterators that can inherently be dereferenced on the device within a SYCL kernel.
```
akukanov, line 15:
```suggestion
Iterators and iterator-like types may or may not refer to content accessible within a `SYCL`_ kernel on a device.
The term *indirectly device accessible* refers to a type that represents content accessible on a device.
An *indirectly device accessible iterator* is such a type that can also be dereferenced within a SYCL kernel.
```

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +362,83 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+
+Customization For User Defined Iterators
+++++++++++++++++++++++++++++++++++++++++
+oneDPL provides a mechanism to indicate whether custom iterators are *indirectly device accessible*.
+
+Applications may define a free function ``is_onedpl_indirectly_device_accessible(T)``, which accepts an argument of
+type ``T`` and returns a type with the base characteristic of ``std::true_type`` if ``T`` is *indirectly device accessible*,
+otherwise returns a type with the base characteristic of ``std::false_type``. The function must be discoverable by
+argument-dependent lookup (ADL). It may be provided as a forward declaration only, without defining a body.
+The return type of ``is_onedpl_indirectly_device_accessible`` is examined at compile time to determine if ``T`` is
```
akukanov, line 387:
```suggestion
argument-dependent lookup (ADL). It may be provided as a forward declaration only, without defining a body.

The return type of ``is_onedpl_indirectly_device_accessible`` is examined at compile time to determine if ``T`` is
```

Looks like in previous suggestions I accidentally removed a space line; better to get it back.

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: I can reduce the number mentioned here, but I think there are probably 4 important categories to call out in this introduction, and I think all are useful to introduce the concept. 

1) iterator with no data (generates on the fly)
2) iterator with device accessible data
3) iterator with only host accessible data
4) buffer wrapper (indirectly device accessible but not iterator)

I could be convinced to drop (1), since we have those on this same page.

akukanov: I tend to think these examples belongs to the developer guide rather than the specification.
(4) is OK as a requirement, not an example. (3) I would also drop, or perhaps combine with (2) - for example, `std::vector` with a USM allocator has an iterator that is indirectly device accessible, bit not with a host allocator.

akukanov: Though maybe `std::vector` with USM allocator is bad for an example, as we do not want to have an impression that it is guaranteed to work efficiently.

danhoeflinger: I cut it down to one positive and one negative example.  If we want to just remove this paragraph we can, and just rely upon the definition and rules below.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -174,6 +221,13 @@ to index into the index map. The corresponding value in the map is then used
 to index into the value set defined by the source iterator. The resulting lvalue is returned
 as the result of the operator.
 
+``permutation_iterator`` is SYCL device-copyable if both the ``SourceIterator`` and the ``IndexMap``
+are SYCL device-copyable. ``permutation_iterator`` is *indirectly device accessible* if both the
+``SourceIterator`` and the ``IndexMap`` are *indirectly device accessible*.
+
+When using ``permutation_iterator`` in combination with an algorithm with a ``device_policy``, ``SourceIterator`` must
+be *indirectly device accessible*.
```
mmichel11, line 229: For our other iterator adaptors, we state that the source iterator must satisfy `AdaptingIteratorSource`. Should we say the same here for `SourceIterator`?

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +361,84 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+
+Once ``is_onedpl_indirectly_device_accessible(T)`` is defined, the public trait
+``template<typename T> oneapi::dpl::is_indirectly_device_accessible[_v]`` will return the appropriate value. This public
+trait can also be used to define the return type of ``is_onedpl_indirectly_device_accessible(T)`` by applying it to any 
+source iterator component types. Refer to the example below.
+
+Example
+^^^^^^^
+
+The following example shows how to define a customization for `is_indirectly_device_accessible` trait for a simple
```
mmichel11, line 400:
```suggestion
The following example shows how to define a customization for ``is_indirectly_device_accessible`` trait for a simple
```
At least in the GitHub preview, this is needed to get the correct formatting.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -301,3 +361,84 @@ operation were applied to each of these iterators. The types ``T`` within the te
...
+
+.. _iterators-device-accessible:
+
+Customization For User Defined Iterators
+++++++++++++++++++++++++++++++++++++++++
+oneDPL provides a mechanism to indicate whether custom iterators are *indirectly device accessible*.
+
+Applications may define a free function ``is_onedpl_indirectly_device_accessible(T)``, which accepts an argument of
+type ``T`` and returns a type with the base characteristic of ``std::true_type`` if ``T`` is *indirectly device accessible*,
+otherwise returns a type with the base characteristic of ``std::false_type``. The function must be discoverable by
```
mmichel11, line 384:
```suggestion
type ``T`` and returns a type with the base characteristic of ``std::true_type`` if ``T`` is *indirectly device accessible*.
Otherwise, it returns a type with the base characteristic of ``std::false_type``. The function must be discoverable by
```
Small grammatical suggestion.

mmichel11 (commented):
I have some small suggestions after rereading the latest version. I am ready to approve once these are considered.

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: We do say that, but it is a few paragraphs above.  It is odd though that there is so much separation between requirements of the same component iterator. I'll try to group them together.

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: Taken.

source/elements/oneDPL/source/parallel_api/iterators.rst
danhoeflinger: thanks, taken.

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -3,9 +3,52 @@
...
+.. code:: cpp
+
+    template <typename T>
+    struct is_indirectly_device_accessible{ /* see below */ };
+
+    template <typename T>
+    inline constexpr bool is_indirectly_device_accessible_v = is_indirectly_device_accessible<T>::value;
+
+``template <typename T> oneapi::dpl::is_indirectly_device_accessible`` is a type which has the base characteristic
+of ``std::true_type`` if ``T`` is *indirectly device accessible*, otherwise it has the base characteristic of
```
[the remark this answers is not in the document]
danhoeflinger: taken

source/elements/oneDPL/source/parallel_api/iterators.rst
mmichel11: I missed it originally. The grouping is a good improvement.

mmichel11 (approved):
LGTM

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -3,9 +3,52 @@
...
+
+When passed to oneDPL algorithms with a ``device_policy``, *indirectly device accessible iterator* types that are also
+random access iterators and satisfy *SYCL device-copyable* must not cause unnecessary data movement beyond what is
+required by the algorithm's semantics and what would be required to use the type directly within a SYCL kernel.
+*Indirectly device accessible* *buffer position objects* must not cause unnecessary data movement beyond what is
+required by the algorithm's semantics and what would be required by using an accessor to the buffer within a SYCL
+kernel.
+
+Public Trait: ``oneapi::dpl::is_indirectly_device_accessible[_v]``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
akukanov, line 32:
```suggestion
Indirect Device Accessibility Trait
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

akukanov (commented):
Below are some formatting comments, after inspecting the rendered HTML page.

Also, to me it seems excessive to italicize every occurrence of the new terms. In principle, it's enough to do once when the term is introduced; see https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.4-rev-1/elements/onedpl/source/parallel_api/buffer_wrappers as an example for *buffer position objects*. I think it is OK to highlight it sometimes though, when the term is sort of reintroduced (perhaps with a link to the original definition).

akukanov (approved):

source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -10,20 +10,41 @@ Iterators
...
+
+.. code:: cpp
+
+    template <typename T>
+    struct is_indirectly_device_accessible{ /* see below */ };
+
+    template <typename T>
+    inline constexpr bool is_indirectly_device_accessible_v = is_indirectly_device_accessible<T>::value;
+
+``template <typename T> oneapi::dpl::is_indirectly_device_accessible`` is a type which has the base characteristic
```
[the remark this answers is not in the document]
danhoeflinger: https://github.com/uxlfoundation/oneAPI-spec/pull/622

PR #621, opened 2025-04-22
[oneMath][RNG] Add PCG64 DXSM generator to Device API

iMartyan:
The PCG is a family of random number generators based on linear congruential generator (LCG). The algorithm applies permutation function to improve statistical properties of LCG.
PCG64 DXSM uses `double xorshift multiply` output function and has a period of 2^128.
Added new page for the engine, modified bibliography and engines overview.

### DIFF
--- a/source/elements/oneMath/source/domains/rng/device_api/device-rng-pcg64-dxsm.rst
+++ b/source/elements/oneMath/source/domains/rng/device_api/device-rng-pcg64-dxsm.rst
@@ -1,0 +1,162 @@
+.. SPDX-FileCopyrightText: 2025 Intel Corporation
+..
+.. SPDX-License-Identifier: CC-BY-4.0
+
+.. _onemath_device_rng_pcg64_dxsm:
+
+pcg64_dxsm
+==========
+
+A permuted congruential pseudorandom number generator PCG64 DXSM with a period of :math:`2^{128}` :ref:`[pcg2014] <onemath_rng_bibliography>`.
+
+.. rubric:: Description
+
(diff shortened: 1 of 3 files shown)


iMartyan:
Formulas:
image

andreyfe1 (approved):
Looks good to me

source/elements/oneMath/source/domains/rng/device_api/device-rng-pcg64-dxsm.rst
```diff
@@ -0,0 +1,162 @@
...
+    .. rubric:: Class Members
+
+    .. list-table::
+        :header-rows: 1
+
+        * - Routine
+          - Description
+        * - `pcg64_dxsm()`_
+          - Default constructor
+        * - `pcg64_dxsm(std::uint64_t seed, std::uint64_t offset = 0)`_
```
aelizaro, line 79: a question about constructors: how big pcg64_dxsm state is? Two questions:
1) whether initial seed should be 32 or 64 bits (other oneMKL engines may have different once)
2) do we need initializer_list versions at all? (e.g. mcg59 https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneMath/source/domains/rng/device_api/device-rng-mcg59.rst has only 64-bit uint state, hence it doesn't need to have extra constructors).

iMartyan: The state of the engine has 128 bits. This way `initializer_list` is needed to fully occupy the state. Same applies to offset as period is 2^128.
About the seed, I suggest to keep it 64 bits, as it will populate half of the state.

aelizaro (approved):

PR #625, opened 2025-05-09
[oneDPL] indirectly_device_accessible adjust hidden friend example 

danhoeflinger:
A body-less hidden friend can look like a common error for friend functions where one intends to externally define the friend, but omits a forward declaration of a template function and template syntax within the class.  

There is no actual problem with the body-less hidden friend, but to avoid looking like this common trap, lets define a minimal body in this case (which creates a one-to-one mapping with the instances of the template class).

This doesn't change the semantics, only our examples and therefore recommendations for usage.

(g++ warns when it sees something like the example prior to this change)

### DIFF
--- a/source/elements/oneDPL/source/parallel_api/iterators.rst
+++ b/source/elements/oneDPL/source/parallel_api/iterators.rst
@@ -429,9 +429,13 @@ of the iterator class.
     {
         It1 first;
         It2 second;
-        friend auto is_onedpl_indirectly_device_accessible(it_pair) ->
+        friend auto
+        is_onedpl_indirectly_device_accessible(it_pair) ->
             std::conjunction<oneapi::dpl::is_indirectly_device_accessible<It1>,
-                             oneapi::dpl::is_indirectly_device_accessible<It2>>;
+                             oneapi::dpl::is_indirectly_device_accessible<It2>>
+        {
+            return {};
+        }
     };
 
     static_assert(oneapi::dpl::is_indirectly_device_accessible<


source/elements/oneDPL/source/parallel_api/iterators.rst
```diff
@@ -429,9 +429,13 @@ of the iterator class.
     {
         It1 first;
         It2 second;
-        friend auto is_onedpl_indirectly_device_accessible(it_pair) ->
+        friend auto
+        is_onedpl_indirectly_device_accessible(it_pair) ->
             std::conjunction<oneapi::dpl::is_indirectly_device_accessible<It1>,
-                             oneapi::dpl::is_indirectly_device_accessible<It2>>;
+                             oneapi::dpl::is_indirectly_device_accessible<It2>>
+        {
```
timmiesmith, line 436: Does this need to be added to the specification for the implementation in https://github.com/uxlfoundation/oneDPL/pull/2241 to remain compliant?  To me the introduction of the body definition in oneDPL is an implementation detail that's being done to address compiler warnings.  Another implementation that didn't provide the definition would still have the same semantic behavior, correct?

danhoeflinger: >Another implementation that didn't provide the definition would still have the same semantic behavior, correct?

Yes, this is how I understand it. We are not changing semantics, just avoiding an example or implementation suggestion for users which may get them to encounter warnings, or the perception of an error. No language needs to change here, its still valid to use a hidden friend, and its still valid to omit the body of the function.  If you do them both at the same time, then you open yourself up to this (false) warning.

timmiesmith: Ok. As this is being provided as an example it looks good to me.

timmiesmith (approved):
LGTM

akukanov (approved):

PR #627, opened 2025-05-28
[oneDPL] Add more parallel range algorithms 

MikeDvorskiy:


akukanov:
Replaced by a series of PRs from @dmitriy-sobolev

PR #635, opened 2025-07-28
[oneDPL] Add more copying mutating parallel range algorithms

dmitriy-sobolev:
Adding `reverse_copy` and `unique_copy` parallel range algorithms into oneDPL specification.

### DIFF
--- a/source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
+++ b/source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
@@ -17,7 +17,7 @@ defined by the C++ standard in ``namespace std::ranges``, they cannot be found b
 and cannot be called with explicitly specified template arguments. [*Note*: A typical implementation uses
 predefined function objects which static function call operators have the required signatures. -- *end note*]
 
-The following differences to the standard C++ serial range algorithms apply:
+The following differences to the standard C++ range algorithms apply:
 
 - Parallel range algorithms cannot be used in constant expressions.
 - The oneDPL execution policy parameter is added.
@@ -32,6 +32,7 @@ The following differences to the standard C++ serial range algorithms apply:
 - For algorithms with bounded output ranges, processing may not need to go over all the input data.
   In that case, the returned value contains iterators pointing to the positions past the last elements
   processed according to the algorithm semantics.
+- ``for_each`` does not return its function object.
 - The return type of ``reverse_copy`` is ``std::ranges::in_in_out_result``
   rather than ``std::ranges::reverse_copy_result``.
   The semantics of the returned value are as specified in
--- a/source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
+++ b/source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
@@ -35,7 +35,7 @@ The following differences to the standard C++ range algorithms apply:
 - ``for_each`` does not return its function object.
 - ``reverse_copy`` returns ``std::ranges::in_in_out_result`` rather than its alias,
   ``std::ranges::reverse_copy_truncated_result``, which is proposed for C++26 in
-  [P3709R2](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3709r2.html).
+  `P3709R2 <https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3709r2.html>`_.
 
 Except for these differences, the signatures of parallel range algorithms correspond to the working draft
 of the next edition of the C++ standard (C++26).
--- a/source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
+++ b/source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
@@ -17,7 +17,7 @@ defined by the C++ standard in ``namespace std::ranges``, they cannot be found b
 and cannot be called with explicitly specified template arguments. [*Note*: A typical implementation uses
 predefined function objects which static function call operators have the required signatures. -- *end note*]
 
-The following differences to the standard C++ range algorithms apply:
+The following differences to the standard C++ serial range algorithms apply:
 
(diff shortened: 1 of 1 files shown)


source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -33,22 +33,29 @@ The following differences to the standard C++ range algorithms apply:
...
 
    // C++20 analogue of std::projected_value_t; exposition only
    template<typename I, typename Proj>
    using /*projected-value-type*/ = std::remove_cvref_t<std::invoke_result_t<Proj&, std::iter_value_t<I>&>>;
 
+  // C++20 analogue of std::ranges::reverse_copy_truncated_result
+  namespace oneapi::dpl::ranges {
+    template<typename I, typename O>
+    using reverse_copy_truncated_result = std::ranges::in_in_out_result<I, I, O>;
+  }
```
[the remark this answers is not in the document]
akukanov: + @rarutyun
Even though it is only an alias, I am not sure if adding it as a public name in the oneDPL namespace is a good idea, given that the same alias will be added into C++26.

I incline towards either making it exposition-only name, like the projected value type, of just using `in_in_out_result`.

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -33,6 +33,8 @@ The following differences to the standard C++ range algorithms apply:
   In that case, the returned value contains iterators pointing to the positions past the last elements
   processed according to the algorithm semantics.
 - ``for_each`` does not return its function object.
+- ``reverse_copy`` returns ``std::ranges::in_in_out_result`` rather than its alias,
+  ``std::ranges::reverse_copy_truncated_result``.
```
akukanov, line 37: Depending on what should be considered as the "base" C++ standard, this wording might be inaccurate. As of now, the rest of the list compares with C++20/23. The new alias for the truncated result of parallel `reverse_copy` will only appear in C++26.

dmitriy-sobolev: I've updated that note in bb4ac307111891e6e2a1dc6126a1c7f80ee03a06 adding more details.

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -33,6 +33,9 @@ The following differences to the standard C++ range algorithms apply:
   In that case, the returned value contains iterators pointing to the positions past the last elements
   processed according to the algorithm semantics.
 - ``for_each`` does not return its function object.
+- ``reverse_copy`` returns ``std::ranges::in_in_out_result`` rather than its alias,
+  ``std::ranges::reverse_copy_truncated_result``, which is proposed for C++26 in
+  `P3709R2 <https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3709r2.html>`_.
```
danhoeflinger, line 38: I think calling it "its alias" may be slightly misleading because `reverse_copy_truncated_result` is not _exactly_ the same as `in_in_out_result`, in that it takes only 2 iterator types and replicates the first twice when using `in_in_out_result`.

```suggestion
- ``reverse_copy`` returns ``std::ranges::in_in_out_result`` rather than the alias,
  ``std::ranges::reverse_copy_truncated_result``, which is proposed for C++26 in
  `P3709R2 <https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3709r2.html>`_.
```

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -458,6 +472,18 @@ Copying Mutating Operations
         transform (ExecutionPolicy&& pol, R1&& r1, R2&& r2, OutR&& result, Fn binary_op,
                    Proj1 proj1 = {}, Proj2 proj2 = {});
 
+    // unique_copy
+    template <typename ExecutionPolicy, std::ranges::random_access_range R,
+              std::ranges::random_access_range OutR, typename Proj = std::identity,
+              std::indirect_equivalence_relation< std::projected<std::ranges::iterator_t<R>, Proj> >
```
danhoeflinger, line 478: I believe this diverged from our normal formatting with extra spaces.

```suggestion
              std::indirect_equivalence_relation<std::projected<std::ranges::iterator_t<R>, Proj>>
```

danhoeflinger (approved):
LGTM

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
dmitriy-sobolev: There were some cases with additional spaces, for example `std::indirect_result_t<Fn&, std::projected<std::ranges::iterator_t<R>, Proj>> >`, but it differs from algorithm to algorithm. I've removed them here.

dmitriy-sobolev:
@danhoeflinger, sorry for dismissing your approval.

I've made a correction. Now, `reverse_copy` interface is compared against the standard serial one, not the parallel API from the draft as it is done with the rest algorithms.

danhoeflinger (approved):
reapproving

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -32,7 +32,10 @@ The following differences to the standard C++ range algorithms apply:
 - For algorithms with bounded output ranges, processing may not need to go over all the input data.
   In that case, the returned value contains iterators pointing to the positions past the last elements
   processed according to the algorithm semantics.
-- ``for_each`` does not return its function object.
+- The return type of ``reverse_copy`` is ``std::ranges::in_in_out_result``
```
akukanov, line 35: The for_each note is removed by mistake, I guess.

PR #648, opened 2025-09-23
[oneDPL][ranges] Add c++23 search algorithms

dmitriy-sobolev:
Adding ``contains``, ``contains_subrange``, ``find_last``, ``find_last_if``, ``find_last_if_not``, ``starts_with`` and ``ends_with`` parallel range algorithms into oneDPL specification.


akukanov (commented):
I think `contains_subrange`, `starts_with` and `ends_with` would be better categorized as sequence search, rather than element search.

PR #650, opened 2025-10-02
[oneTBB] Add a function to create a set of NUMA bound task arenas

isaevil:
Add a page that describes helper APIs to simplify constrained arena creations. Currently it is a single function `oneapi::tbb::create_numa_task_arenas` from [corresponding RFC](https://github.com/uxlfoundation/oneTBB/blob/981eb42aee3730ced4c296e16a3154e1803bd088/rfcs/proposed/numa_support/create-numa-arenas.md).

### DIFF
--- a/source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
+++ b/source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
@@ -376,7 +376,7 @@ to the corresponding NUMA node.
         }
 
         arenas[0].execute([] {
-            /* parallel work */
+            /* executed by the main thread pinned to the NUMA node for arenas[0] */
         });
 
         for (int i = 1; i < arenas.size(); i++) {


isaevil:
@aleksei-fedotov @akukanov Please take a look.

source/elements/oneTBB/source/constrained_arena_helpers.rst
```diff
@@ -0,0 +1,38 @@
...
+        std::vector<oneapi::tbb::task_arena> create_numa_task_arenas(
+            oneapi::tbb::task_arena::constraints other_constraints = {},
+            unsigned reserved_slots = 0);
+    } // namespace tbb
+    } // namespace oneapi
+
+Functions
+---------
+
+.. cpp:function:: std::vector<oneapi::tbb::task_arena> create_numa_task_arenas(oneapi::tbb::task_arena::constraints other_constraints = {}, unsigned reserved_slots = 0)
```
aleksei-fedotov, line 27: Shorten?
```suggestion
.. cpp:function:: std::vector<oneapi::tbb::task_arena> create_numa_task_arenas(oneapi::tbb::task_arena::constraints constraints = {}, unsigned reserved_slots = 0)
```

source/elements/oneTBB/source/constrained_arena_helpers.rst
```diff
@@ -0,0 +1,38 @@
...
+
+    Returns a ``std::vector`` of ``task_arena`` objects, each bound to a separate NUMA node.
+    The number of created ``task_arena`` is equal to the number of NUMA nodes available on the system.
+    It optionally takes additional ``other_constraints`` argument to be applied to each created
+    ``task_arena``, ignoring ``numa_id`` value. The second optional ``reserved_slots`` argument
+    allows to reserve a number of slots in each created ``task_arena`` for application threads.
```
aleksei-fedotov, line 33: Perhaps,
```suggestion
    Additional ``constraints`` argument can be specified to further limit the threads joined
    the ``task_arena`` objects. The ``numa_id`` value in the ``constraints`` argument is
    ignored. The ``reserved_slots`` argument allows reserving specified number of slots in
    ``task_arena`` objects for application threads.
```

source/elements/oneTBB/source/constrained_arena_helpers.rst
```diff
@@ -0,0 +1,38 @@
...
+            oneapi::tbb::task_arena::constraints other_constraints = {},
+            unsigned reserved_slots = 0);
+    } // namespace tbb
+    } // namespace oneapi
+
+Functions
+---------
+
+.. cpp:function:: std::vector<oneapi::tbb::task_arena> create_numa_task_arenas(oneapi::tbb::task_arena::constraints other_constraints = {}, unsigned reserved_slots = 0)
+
```
akukanov: As far as I can tell, the specification does not use `oneapi::tbb` in the signatures of functions that are already in that namespace.

isaevil: Removed `oneapi::tbb`

source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
```diff
@@ -343,14 +362,9 @@ to the corresponding NUMA node.
     #include <vector>
 
     int main() {
-        std::vector<oneapi::tbb::numa_node_id> numa_nodes = oneapi::tbb::info::numa_nodes();
-        std::vector<oneapi::tbb::task_arena> arenas(numa_nodes.size());
+        std::vector<oneapi::tbb::task_arena> arenas = oneapi::tbb::create_numa_task_arenas();
         std::vector<oneapi::tbb::task_group> task_groups(numa_nodes.size());
```
akukanov: Since `numa_nodes` were removed, make sure these are not used in the remaining code.

source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
```diff
@@ -329,6 +332,23 @@ Member functions
 
     The behavior of this function is equivalent to ``this->execute([&tg]{ return tg.wait(); })``.
 
+Non-member Functions
+--------------------
+
+.. cpp:function:: std::vector<task_arena> create_numa_task_arenas(task_arena::constraints constraints = {}, unsigned reserved_slots = 0)
+
+    Returns a ``std::vector`` of non-initialized ``task_arena`` objects, each bound to a separate NUMA node.
+    The number of created ``task_arena`` is equal to the number of NUMA nodes on the system,
```
akukanov:
```suggestion
    Returns a ``std::vector`` of uninitialized ``task_arena`` objects, each to be bound to a separate NUMA node.
    The number of created ``task_arena``s is equal to the number of NUMA nodes on the system,
```

source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
```diff
@@ -329,6 +332,23 @@ Member functions
...
+    as determined by ``tbb::info::numa_nodes()``. If an error occurs during system information discovery,
+    returns a ``std::vector`` containing a single ``task_arena`` object created as ``task_arena(constraints, reserved_slots)``.
+
+    The ``constraints`` argument can be specified to further limit threads joining
+    the ``task_arena`` objects. The ``numa_id`` value in ``constraints`` is automatically
+    replaced with a separate value for each created arena.
```
kboyarinov: The second sentence is not 100% clear for me. 
As far as I understand, the `numa_id` value in `constraints` is ignored and the corresponding ``numa_id` from ``tbb::info::numa_nodes`` is used instead.

May be something like this will be easier to understand?

```suggestion
    the ``task_arena`` objects. The ``numa_id`` value specified in ``constraints`` is disregarded, and instead,
    the corresponding ``numa_id`` from ``tbb::info::numa_nodes()`` is used during construction
    of each particular ``task_arena``.
```

Also, should we specify this somehow if the error occurs during the system parsing as part of `create_numa_task_arenas()`? Will the `numa_id` be ignored in this case? 

I understand that if the error occurs in `create_numa_task_arenas`, then it is hard to imagine that there is a valid `numa_id` in constrains, but may be it still make sense to specify the behavior explicitly somehow.

isaevil: Applied suggestions from here and in the comment above from the different discussion.

source/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.rst
```diff
@@ -343,21 +366,16 @@ to the corresponding NUMA node.
...
 
-        for (int i = 0; i < numa_nodes.size(); i++) {
+        for (int i = 0; i < arenas.size(); i++) {
             arenas[i].enqueue([]{
                 /* executed by a thread pinned to the specified NUMA node */
             }, task_groups[i]);
         }
 
-        for (int i = 0; i < numa_nodes.size(); i++) {
+        for (int i = 0; i < arenas.size(); i++) {
```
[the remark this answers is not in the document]
isaevil: That's right. I just didn't want to break that 1:1 match of task_groups to task_arenas. Otherwise the example becomes the following:
```cpp
#include "oneapi/tbb/task_group.h"
#include "oneapi/tbb/task_arena.h"

#include <vector>

int main() {
    std::vector<oneapi::tbb::task_arena> arenas = oneapi::tbb::create_numa_task_arenas();
    std::vector<oneapi::tbb::task_group> task_groups(arenas.size()-1);

    for (int i = 1; i < arenas.size(); i++) {
        arenas[i].enqueue([]{
            /* executed by a thread pinned to the specified NUMA node */
        }, task_groups[i-1]);
    }

    arenas[0].execute([] {
        /* parallel work */
    });

    for (int i = 1; i < arenas.size(); i++) {
        arenas[i].wait_for(task_groups[i-1]);
    }

    return 0;
}
```
Or simply still create equal number of `task_group` and `task_arena` but never use first one in example:
```cpp
#include "oneapi/tbb/task_group.h"
#include "oneapi/tbb/task_arena.h"

#include <vector>

int main() {
    std::vector<oneapi::tbb::task_arena> arenas = oneapi::tbb::create_numa_task_arenas();
    std::vector<oneapi::tbb::task_group> task_groups(arenas.size());

    for (int i = 1; i < arenas.size(); i++) {
        arenas[i].enqueue([]{
            /* executed by a thread pinned to the specified NUMA node */
        }, task_groups[i]);
    }

    arenas[0].execute([] {
        /* parallel work */
    });

    for (int i = 1; i < arenas.size(); i++) {
        arenas[i].wait_for(task_groups[i]);
    }

    return 0;
}
```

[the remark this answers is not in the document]
akukanov: Either seems good enough to me.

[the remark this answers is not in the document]
vossmjp: I suppose there's no particular reason to assume that arena[0] is the one that should be executed in. So you could create 1 less arena, enqueue into the first size()-1 and then execute in the last arena. Perhaps then the first option looks slightly better not needing to mismatch the indexes when calling enqueue. But I'd also be ok with either of the suggested examples as they are.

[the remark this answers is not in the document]
isaevil: I have updated the example.

PR #658, opened 2025-10-28
[oneDPL] Fix the default template argument for the new value type in replace[_if]

akukanov:
Reviewers of the C++26 working draft have found that, citing,
> The default template argument for the type of the new value in `range::replace` and `ranges::replace_if` should not have projections applied.

While this comment has not yet been resolved, it seems correct: the standard wording for these algorithms does not imply that the new value can be used as an argument to the projection or in an expression with projected values.

It seems appropriate therefore to proactively fix the respective signatures in the oneDPL specification.

### DIFF
--- a/source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
+++ b/source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
@@ -868,7 +868,8 @@ In-place Mutating Operations
     // replace
     template <typename ExecutionPolicy, std::ranges::random_access_range R,
               typename Proj = std::identity,
-              typename T1 = /*projected-value-type*/<std::ranges::iterator_t<R>, Proj>, typename T2 = T1>
+              typename T1 = /*projected-value-type*/<std::ranges::iterator_t<R>, Proj>,
+              typename T2 = std::ranges::range_value_t<R>>
       requires oneapi::dpl::is_execution_policy_v<std::remove_cvref_t<ExecutionPolicy>> &&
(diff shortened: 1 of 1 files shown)


rarutyun (approved):
The interesting part of this PR is that we don't have any tests that fail but that was something we have discussed offline already

danhoeflinger (approved):

source/elements/oneDPL/source/parallel_api/parallel_range_api.rst
```diff
@@ -868,7 +868,8 @@ In-place Mutating Operations
     // replace
     template <typename ExecutionPolicy, std::ranges::random_access_range R,
               typename Proj = std::identity,
-              typename T1 = /*projected-value-type*/<std::ranges::iterator_t<R>, Proj>, typename T2 = T1>
+              typename T1 = /*projected-value-type*/<std::ranges::iterator_t<R>, Proj>,
+              typename T2 = std::ranges::range_value_t<R>>
```
dmitriy-sobolev: This asymmetry between `old_value` and `new_value` seems odd at first, given their names.

However, I understand that this asymmetry is necessary. It's like replacing by key. We specify a key (projection applied), `old_value`, and we specify a key-value pair, `new_value`, to replace a found entry with.

I agree that we should act proactively. But, I would anticipate more changes in the standard, e.g. different names of the arguments, which is not crucial.

dmitriy-sobolev (approved):

PR #667, opened 2026-03-16
[oneMath][BLAS] Deprecate 'gemmt' and rename it to 'gemmtr'

dnhsieh-intel:
This PR deprecates oneMath BLAS-like extension `gemmt` and introduces BLAS level 3 `gemmtr` as its new name.


andrewtbarker (approved):
Looks good to me!

